PR #21, opened 2021-03-19
Feature/45711 news page updates

Sean-Stacey:
Not sure about the formatting on this one, the layout looks a bit wrong.  Any suggestions?

### DIFF
--- a/Web/Edubase.Web.UI/Views/Home/News.cshtml
+++ b/Web/Edubase.Web.UI/Views/Home/News.cshtml
@@ -20,10 +20,11 @@
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
         <h1 class="govuk-heading-xl govuk-!-margin-bottom-0">News</h1>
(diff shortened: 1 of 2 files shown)


jonbanner:
@Sean-Stacey I'd suggest removing the <br> tags.
Probably a good idea to standardise the govuk margin utility classes between the 2 pages too. (and use the one's on the header to create the gap you are after)

the govuk-!-margin-bottom-<int> classes use the govuk-spacing helper to provide responsive marging - documentation here: https://design-system.service.gov.uk/styles/spacing/

Sean-Stacey:
Cheers, will take a look.

jonbanner:
@Sean-Stacey not sure that you want the class names on the paragraph tags.

The default styling for the paragraphs would align exactly with the govuk-body class, so that is redundant on the main news page.

I'd defer to Kevin on this, but i would think that over riding the the bottom margin on the paragraph tags would also look a bit strange when comparing to the rest of the paragraphs in the page as they are using the GOVUK default styling. 

Alternatively you could merge the 2 paragraphs and use a `<br>` to force the required line break.

PR #23, opened 2021-03-25
Feature/45711 news page updates

Sean-Stacey:
New pull request raised to address target; also pulled class from paras.


jonbanner:
looks good to me @Sean-Stacey

PR #51, opened 2021-05-10
Updated spacing and heading width; viewport seems OK as per AC

Sean-Stacey:


jonbanner:
@Sean-Stacey I'm not sure that we should allow xl size headings in page - they would look very strange and should only be used on the main page heading. 
Maybe one to discuss with Kevin when he's around, or to at least get some more clarity around the expected font sizes for the headings at desktop and mobile. (the govuk heading classes change font size depending on the screen size)
I think there was a requirement to wrap the content to 2/3rds width on desktop so you'd also need to update the container class from `.govuk-grid-column-full` to `.govuk-grid-column-two-thirds`

Sean-Stacey:
Story on VSO has additional comments from Kevin re. heading styles.

PR #54, opened 2021-05-13
Feature/46831 remove dropdown

Sean-Stacey:
Looks like this removes the dropdown and correctly sets titlesize


jonbanner:
that ought to take care of it @Sean-Stacey

PR #56, opened 2021-05-19
44167 - Think this covers all ACs other than AC9 which I understand i…

Sean-Stacey:
…s going to be raised as a separate item.


jonbanner:
looks ok to me @Sean-Stacey. 

There's a lot of changes in the group details view - looks like whitespace changes, but i'd suggest a quick once over between your local and dev just to make sure you've not changed anything you didn't intend.

Sean-Stacey:
Yeah, that doesn't look right.  There's a lot of changes there that I've had nothing to do with (ca. line 200, I've not touched any of that stuff).  I'll take a copy of my changes, bin this and start again from a fresh pull from dev.

PR #57, opened 2021-05-20
44167 - redo of breadcrumb changes incorporating recent updates

Sean-Stacey:
Still a large number of mods in group details file - the only change was from section backlink to breadcrumb.  Windiff suggests that yes, these are just whitespace changes.


jonbanner:
@Sean-Stacey changes look ok - like you say just white space and if not, I'm sure Mattias will spot it.

PR #59, opened 2021-05-25
Updates, removing 'back to search results' x 4

Sean-Stacey:
44167:related - made changes to four additional pages in line with the intent.


jonbanner:
looks good to go @Sean-Stacey

PR #60, opened 2021-05-25
46938-correction-to-copy

Sean-Stacey:
44167-change of text as per Osman/Kevin's comments.


jonbanner:
looks appropriate @Sean-Stacey

PR #70, opened 2021-06-17
Found a couple of additional interstital pages I'd not noticed previo…

Sean-Stacey:
…usly.

Pages are additional content on the Establishment downloads journey that were located in a different section of the site and so were missed.  The update is the same section as the 'Establishments' component of the earlier pages (and reformatting of whitespace) Ref. bug 47228.


jonbanner:
@Sean-Stacey looks ok - the commit message needs amending to include the ticket number and perhaps a better description of what has been changed or why. 

If you run `git commit --amend` you'll be able to update the commit message locally, then push

PR #72, opened 2021-06-22
47228-bugfix-return-to-tab-on-index-page

Sean-Stacey:
Modification to eight pages.  "Search" breadcrumb now redirects to correct tab on index page.


jonbanner:
hi @Sean-Stacey i'm in meetings all afternoon - but at a quick scan, should the selectedTab variable be hard coded to establishments? And you are rendering breadcrumbs in the back link slot, which doesn't break anything as they are adjacent- i think you should change to the breadcrumbs then once all the backlinks have been replaced we can remove the backlink slot from the template.

Web/Edubase.Web.UI/Areas/Establishments/Views/EstablishmentsSearch/Downloads/SelectDataset.cshtml
```diff
@@ -24,7 +24,7 @@
             <div class="govuk-grid-column-full">
                 <div class="govuk-breadcrumbs">
                     <ol class="govuk-breadcrumbs__list">
-                        <li class="govuk-breadcrumbs__list-item">@Html.ActionLink("Search", "Index", "Search", new { area = "" }, new { @class = "govuk-breadcrumbs__link" })</li>
+                        <li class="govuk-breadcrumbs__list-item">@Html.ActionLink("Search", "Index", "Search", new { area = "", SelectedTab = "Establishments" }, new { @class = "govuk-breadcrumbs__link" })</li>
```
MEvely, line 27: SelectedTab should possibly use the Model.SearchSource (referenced in line 17) in a similar manner to that used when building the link on EstablishmentSearch/Index.cshtml

hard-coding might be valid if this is only ever going to be driven by the establishment search, in which case it might be worth using the SearchViewModel.Tab.Establishments enum ref rather than a string as ensures consistency.

Web/Edubase.Web.UI/Views/Downloads/PreparingFilePleaseWait.cshtml
```diff
@@ -24,7 +24,7 @@
         <div class="govuk-grid-column-full">
             <div class="govuk-breadcrumbs">
                 <ol class="govuk-breadcrumbs__list">
-                    <li class="govuk-breadcrumbs__list-item">@Html.ActionLink("Search", "Index", "Search", new { area = "" }, new { @class = "govuk-breadcrumbs__link"})</li>
+                    <li class="govuk-breadcrumbs__list-item">@Html.ActionLink("Search", "Index", "Search", new { area = "" }, new { @class = "govuk-breadcrumbs__link" })</li>
```
MEvely, line 27: looks like this file could be excluded from the commit as it was only to introduce whitespace so cleaner to keep the commits as isolated as possible for files that were impacted

PR #78, opened 2021-07-19
47417/47656 - Changes for footer relating to privacy statement

Sean-Stacey:
Updated static pages (just) as well as template with privacy notice.  Some autoformatting changes also applied.

### DIFF
--- a/Web/Edubase.Web.UI/403.aspx
+++ b/Web/Edubase.Web.UI/403.aspx
@@ -1,4 +1,5 @@
 <%@ Page Language="C#" %>
+
 <% Response.StatusCode = 404; %>
 
 <!DOCTYPE html>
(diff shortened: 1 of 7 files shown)


Web/Edubase.Web.UI/403.aspx
```diff
@@ -212,11 +213,13 @@
                 <div class="footer-meta">
                     <div class="footer-meta-inner">
                         <h2 class="visuallyhidden">Support links</h2>
-                        <ul>
-                            <li><a href="/cookies">Cookies</a></li>
-                            <!--<li><a href="#">Contact</a></li>-->
-                            <li><a href="/guidance">Guidance</a></li>
-                            <li><a href="/news">News</a></li>
+                        <ul class="govuk-footer__inline-list">
+                            <li class="govuk-footer__inline-list-item"><a href="https://form.education.gov.uk/service/Data-collections-service-request-form" target="_blank" rel="noreferrer noopener" class="govuk-footer__link">Contact (opens in new tab)</a></li>
```
MEvely, line 217: Did you mean to update the Contact link as well, as dont think this was expected as part of this story (and the general page footer directs users to the contact page not needing to open in a new window)

Web/Edubase.Web.UI/Views/GOVUK/Layouts/template.cshtml
```diff
@@ -67,222 +67,223 @@
...
+            <div class="govuk-width-container">
+                <div class="govuk-footer__meta">
+                    <div class="govuk-footer__meta-item govuk-footer__meta-item--grow">
+                        <h2 class="govuk-visually-hidden">Support links</h2>
+                        <ul class="govuk-footer__inline-list">
+                            <li class="govuk-footer__inline-list-item">@Html.ActionLink("Contact", "Contact", "Home", new { area = "" }, new { @class = "govuk-footer__link" })</li>
+                            <li class="govuk-footer__inline-list-item">@Html.ActionLink("Cookies", "Cookies", "Home", new { area = "" }, new { @class = "govuk-footer__link" })</li>
+                            <li class="govuk-footer__inline-list-item">@Html.RouteLink("Glossary", "Glossary", null, new { @class = "govuk-footer__link" })</li>
+                            <li class="govuk-footer__inline-list-item">@Html.ActionLink("Guidance", "Index", "Guidance", new { area = "" }, new { @class = "govuk-footer__link" })</li>
+                            <li class="govuk-footer__inline-list-item"><a class = "govuk-footer__link" href="\Privacy">Privacy</a></li>
```
MEvely, line 260: Think this should be using an ActionLink/RouteLink in a similar manner to the links above

MEvely (changes requested):
Looks like the contact link text has been updated on the static 4xx/5xx aspx/html files too, but that does not look to use the same link as the normal page footer.

Also got to be really careful about making wide range reformat changes on the pages as it makes it really difficult to see the actual underlying change that had been introduced when doing a comparison for these PRs

PR #85, opened 2021-08-10
47786-47458 Removing back arrows for four pages: EstablishmentSearch …

Sean-Stacey:
…/ GroupSearch / GovernorSearch / DownloadError

### DIFF
--- a/Web/Edubase.Web.UI/Views/Shared/Downloads/DownloadError.cshtml
+++ b/Web/Edubase.Web.UI/Views/Shared/Downloads/DownloadError.cshtml
@@ -36,7 +36,15 @@
 {
     @if (Model.SearchQueryString.IsNullOrWhiteSpace() || Model.SearchSource == null)
     {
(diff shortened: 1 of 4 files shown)


Web/Edubase.Web.UI/Views/Shared/Downloads/DownloadError.cshtml
```diff
@@ -36,7 +36,7 @@
 {
     @if (Model.SearchQueryString.IsNullOrWhiteSpace() || Model.SearchSource == null)
     {
-        @Html.ActionLink("Search", "Index", "Search", new { area = "" }, new { @class = "govuk-back-link" })
+        @Html.ActionLink("Search", "Index", "Search", new { area = "" }, new { @class = "govuk-breadcrumbs__link" })
```
MEvely, line 39: In this scenario I think we'll want to include the divs (similar to the other links within the ELSE statement).

An easy way to test what this renders like, is to go to the main downloads page, select a small filesize file, and then click through until you download it. If you click on the Results.zip button a second time it will show this error page with the empty Search link at the top.

MEvely (approved):

PR #88, opened 2021-08-18
48121 styling on downloads

Sean-Stacey:


jonbanner:
@Sean-Stacey that doesn't look quite right to me.

have a look at https://design-system.service.gov.uk/components/breadcrumbs/
the list items should have the class _govuk-breadcrumbs__list-item_
and the links themselves need the class _govuk-breadcrumbs__link_

To add a class using the ActionLink helper you'll likely need something along these lines

`@Html.ActionLink("LinkText", "ActionName", "ControllerName", new {area = "Establishments"}, new {@class = "govuk-breadcrumbs__link"})`

there is an example of the breadcrumbs in \Edubase.Web.UI\Areas\Establishments\Views\BulkUpdate\InProgress.cshtml


drop me a line on slack if this doesn't make sense

Sean-Stacey:
@jonbanner Have amended my approach, can you look over it and let me know if it makes sense now?

jonbanner:
That looks right to me @Sean-Stacey - merge it in and see what Mattias thinks when he gets back from his break.

PR #100, opened 2021-10-06
Changes to styling on pages 2-3 on non-Children's Centre route

Sean-Stacey:
Updated styling on pages 2-3 for non-childrens centre route (essentially removing two nested div elements)


superfurryadamal (changes requested):
for these i would keep the div containers but remove the class. The only reason i say this is that those div IDs are probably used in the javascript somewhere. It may be useful to have them in place when we come to re-develop the JS journey

superfurryadamal (approved):

PR #102, opened 2021-10-07
44445 non js journey

Sean-Stacey:


Web/Edubase.Web.UI/Assets/Scripts/Entry/create-establishment.js
```diff
@@ -1,3 +1,7 @@
+import Vue from 'vue';
+const ceApp = new Vue({
+  el: '#create-establishment'
+});
```
superfurryadamal, line 4: we should removed this actually as we are no longer using Vue cloak to show/hide the JS version

PR #109, opened 2021-11-04
49370 - amended fsm tooltip and added fsm percentage tooltip

Adam-Millard:
Added new tooltip for %
Changed wording old existing tooltip


Sean-Stacey:
Looks good to me @Adam-Millard.

Sean-Stacey (approved):

PR #110, opened 2021-11-04
49440 back links

MEvely:
Updated journey to allow backlinks to take user to previous screens rather than all the way out to the tools page. To ensure this worked as required there are now three components which get updated with each page build to reference the previous step, the active (current) step, and the point in the journey the user should be directed to upon a successful postback.

To ensure that there no obsolete fields remain populated if the user changes their mind mid-journey, there are also some conditional elements around the hidden fields depending on at which stage the user is at.


Sean-Stacey (approved):

Adam-Millard (approved):
looks good. Agree with changing the step names from "stepx" to the more meaningful names that describe the action. more self-documenting this way

PR #113, opened 2021-11-19
48703 - fixed openonly chechbox preference for LA disambiguation

Adam-Millard:


MEvely (approved):
good choice with the helpers, make it much cleaner to look through the markup.

PR #120, opened 2021-12-10
Updated Home\Responsibilities.cshtml

Sean-Stacey:


Adam-Millard (changes requested):
Hi Sean

Ticket says: Identify all areas where the 'Academies Financial Handbook' is referenced and replace it with '**the Academy Trust Handbook**'.

Might need to double check with claire but i'd go with how its been written in the ACs

Sean-Stacey:
OK, have to leave it for Monday as Claire's on her non-working day today.  NP.

PR #128, opened 2021-12-31
50809 for consideration

Adam-Millard:
PR for sean's code cleanup work which was completed before he left the team


Adam-Millard (commented):
checked Sean's code tidy commit. 

Added one comment as i notice that the code might need looking at in a future ticket. However, as these changes shouldn't be making any changes to the functionality, happy for this to be merged in

PR #145, opened 2022-01-26
Feature/50859 update independent school types fields

BarryGibney:


Adam-Millard (approved):
as discussed with @BarryGibney

PR #156, opened 2022-03-24
Feature/51806 establishment read api service unit tests

morrisscd:
To add these Unit Tests I have had to introduce a IHttpClientWrapper interface. The issue with this is that I have had to register this in IocConfig.cs. Ideally I should go through and update all the existing references to use this interface, but I have registered it alongside the current HttpClientWrapper registration eg.

    builder.RegisterType<HttpClientWrapper>().AsSelf();
    **builder.RegisterType<HttpClientWrapper>().As<IHttpClientWrapper>();**

I have run the service and this seems ok, but I'm not entirely sure this is the correct thing to do.

Also,  many of the methods in establishmentReadApiService.cs use the injected _httpClient. If the response from this client fails then a general exception is thrown. This means that the branch code coverage reporting for is not picking up these exceptions as covered failure paths. However, the line coverage is 86%.


Web/Edubase.Services.TexunaUnitTests/Establishments/EstablishmentReadApiServiceTests.cs
```diff
@@ -0,0 +1,714 @@
...
+        {
+            //Arrange
+            var apiResponse = new ApiResponse<BoolResult>(true) { Response = new BoolResult() { Value = true } };
+
+            _mockHttpClientWrapper.Setup(x => x.GetAsync<BoolResult>(It.IsAny<string>(), It.IsAny<IPrincipal>()))
+                .ReturnsAsync(() => apiResponse);
+
+            //Act
+            var service = GetEstablishmentReadApiService();
+            var result = await service.CanAccess(It.IsAny<int>(), null);
```
ScottDawsonDfE, line 38: Don't use the It.IsAny<int> here - this is just for Mock Objects

ScottDawsonDfE: Also lines 56, 91

morrisscd: @ScottDawsonDfE ok, updated

ScottDawsonDfE (approved):
Approved

MEvely (approved):
All looks good with this Chris, but noticed the new IHttpClientWrapper being introduced, I thought it could be worth mentioning the api-v2 branch (where we have provided a secondary httpClient connection to the new api) has introduced a new base class for the HttpClientWrapper, so both the traditional api connection and the new one can all share the same base methods.

This obviously isn't an issue with this PR, but thought it could be worth mentioning here as a note for future development, as the api-v2 branch may require a minimal amount of refactoring to inherit the IHttpClientWrapper from the base class rather than the wrapper directly.

morrisscd:
@MEvely Thanks, I'll keep the api-v2 branch you mentioned above in mind

PR #157, opened 2022-03-29
Api/51806 unit test incorporated into api v2

MEvely:
Following the code changes that had been introduced for the TexunaUnitTests project, this combines the code changes for 51806 as well as the work already done in the api-v2 branch (which has the secondary httpClient link to the new api address).

Because all of the Services.Texuna classes had been restructured to include a new ApiClientWrapper class, with the shared methods under a new base class referenced from both the ApiClientWrapper and HttpClientWrapper, the interfaces that had been introduced as part of 51806 needed to be ported to these new base classes.

This will bring the 51806 work which will shortly be merged to dev, inline with the new api-v2 branch, ensuring that there should be no further merge-conflicts with this.

The api-v2 branch is expected to become the dev branch shortly so this should help with that migration.


morrisscd (approved):
looks good, cheers, Chris

esde84 (approved):

PR #185, opened 2022-06-23
Hotfix for 53520 update of NewtonSoft.Json package

esde84:


ScottDawsonDfE:
Ignoring GitLEaks fail as this is due to a change in how gitleaks works. Code has been checked to ensure no changes have introduced any secret information

PR #186, opened 2022-06-23
Newtonsoft hotfix merge from Master to Dev

ScottDawsonDfE:


ScottDawsonDfE:
Ignoring GitLeaks issues due to change in how gitleaks works and now requires a licence key. Code has been checked to ensure no private data has been added

PR #198, opened 2022-08-11
corrected alignment and adjust media query

morrisscd:


morrisscd:
Need to exclude SCSS files from this PR

PR #200, opened 2022-08-11
Reverting Microsoft.AspNet.WebApi.Client from 5.2.9.0 to 5.2.3.0

morrisscd:
Reverting _Microsoft.AspNet.WebApi.Client_ from 5.2.9.0 to 5.2.3.0


morrisscd:
There's a conflict between version 5.2.3.0 and 5.2.9.0 which is causing an error. This need further investigation as it's not clear where version 5.2.3.0 is being used.

ScottDawsonDfE (approved):

PR #225, opened 2022-10-17
Removed use of Redis cache to avoid confusion 

esde84:
and prevent issues following package update.


RogerHowellDfE:
@ScottDawsonDfE - Is this change still required / desirable?

PR #252, opened 2022-12-15
Updated Handling of Glossary Content

morrisscd:
Small change to how content is handled in the Glossary accordion sections

### Currently, new lines are not being shown
image

### With update
image

This correction has already been use in FAQ sections

### DIFF
--- a/Web/Edubase.Web.UI/Views/Glossary/Index.cshtml
+++ b/Web/Edubase.Web.UI/Views/Glossary/Index.cshtml
@@ -49,7 +49,7 @@
                             <div class="govuk-grid-row">
                                 <div class="govuk-grid-column-two-thirds govuk-!-padding-bottom-2">
                                     <h2 class="govuk-heading-s">@item.Title</h2>
-                                    <div class="gias-user-content">@Html.Raw(item.Content)</div>
+                                    <div class="gias-user-content">@Html.Raw(item.Content.Replace("\n", "<br/>"))</div>
                                 </div>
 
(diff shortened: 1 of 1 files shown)


Web/Edubase.Web.UI/Views/Glossary/Index.cshtml
```diff
@@ -49,7 +49,7 @@
                             <div class="govuk-grid-row">
                                 <div class="govuk-grid-column-two-thirds govuk-!-padding-bottom-2">
                                     <h2 class="govuk-heading-s">@item.Title</h2>
-                                    <div class="gias-user-content">@Html.Raw(item.Content)</div>
+                                    <div class="gias-user-content">@Html.Raw(item.Content.Replace("\n", "<br/>"))</div>
```
RogerHowellDfE, line 52: Are newlines within `item.Content` guaranteed to only ever be `\n`?

Consider this more general alternative which matches on any of CRLF / CR / LF (generally these are Windows/Mac/Linux linebreaks):

```csharp
 <div class="gias-user-content">@Html.Raw(Regex.Replace(item.Content, @"\r\n|\r|\n", "<br/>"))</div>

```

https://dotnetfiddle.net/uk1oaH

anthony9853: I agree with Roger.

morrisscd: @RogerHowellDfE Done, thanks for the suggestion

@anthony9853

PR #253, opened 2022-12-28
Sd/resolving ide warnings

ScottDawsonDfE:


Web/Edubase.Services.Texuna/Lookup/LookupApiService.cs
```diff
@@ -146,303 +146,303 @@ public async Task<IEnumerable<EstablishmentLookupDto>> EstablishmentTypesGetAllA
...
+        //
+        // public IEnumerable<LookupDto> AdministrativeWardsGetAll()
+        // {
+        //     throw new NotImplementedException();
+        // }
+        //
+        // public IEnumerable<LookupDto> AdmissionsPoliciesGetAll()
+        // {
+        //     throw new NotImplementedException();
+        // }
```
RogerHowellDfE, line 174: Better to delete these rather than just comment out?

Web/Edubase.Services.TexunaUnitTests/Lookup/LookupApiServiceTests.cs
```diff
@@ -67,16 +67,6 @@ public async Task EstablishmentTypesGetAllAsyncTest()
...
-        [Theory]
-        [MemberData(nameof(GetSyncronousLookupTypes))]
-        public void SyncronousLookupsTest(string lookupType)
-        {
-            //use reflection to construct method name
-            var method = _lookupApiService.GetType().GetMethod($"{lookupType}GetAll");
-            var ex = Assert.Throws<System.Reflection.TargetInvocationException>(() => method.Invoke(_lookupApiService, null));
-            Assert.IsType<NotImplementedException>(ex.InnerException);
-        }
-
```
RogerHowellDfE, line 79: Note to self -- this test is removed because the synchronous (`NotImplementedException`) methods were removed (commented out) from `LookupApiService.cs`

Web/Edubase.Services/Lookup/CachedLookupService.cs
```diff
@@ -94,147 +94,72 @@ public CachedLookupService(ILookupService lookupService, ICacheAccessor cacheAcc
...
                     }
-                    
+
                     return displayName;
                 })) }
             };
         }
 
         public async Task<IEnumerable<LookupDto>> LocalAuthorityGetAllAsync() => await AutoAsync(_lookupService.LocalAuthorityGetAllAsync);
-        public IEnumerable<LookupDto> LocalAuthorityGetAll() => Auto(_lookupService.LocalAuthorityGetAll);
```
RogerHowellDfE, line 104: Note to self -- these lines are removed because the synchronous (`NotImplementedException`) methods were removed (commented out) from `LookupApiService.cs`

ScottDawsonDfE: I should have made a note of this

Web/Edubase.Services.Texuna/Lookup/LookupApiService.cs
ScottDawsonDfE: Yes I should

Web/Edubase.Services.Texuna/ChangeHistory/ChangeHistoryService.cs
```diff
@@ -28,13 +28,13 @@ public async Task<List<UserGroupModel>> GetSuggesterGroupsAsync(IPrincipal princ
         [Obsolete("Please use GetSuggesterGroupsAsync", true)]
         public async Task<List<UserGroupModel>> GetApproversGroupsAsync(IPrincipal principal)
         {
-            throw new NotImplementedException("GetApproversGroupsAsync is not implemented because it returns the exact same information as GetSuggesterGroupsAsync");
+            return await GetSuggesterGroupsAsync(principal);
         }
```
RogerHowellDfE, line 32: Possibly outside the scope of this but if we're modifying this anyway could we just inline this method? Or remove it? 

It appears to have been marked as obsolete for a while and given that it previously threw an exception this suggests it's not used anywhere, so perhaps deletion is the better option.

Web/Edubase.Web.UI/Areas/Establishments/Views/EstablishmentsSearch/Downloads/SelectCustomFields.cshtml
```diff
@@ -9,26 +9,37 @@
...
+                                new { area = "" }, new { @class = "govuk-breadcrumbs__link" })
+                        </li>
+                        <li class="govuk-breadcrumbs__list-item">
+                            @Html.ActionLink("Search", "Index", "Search",
+                                new { area = "", SelectedTab = SearchViewModel.Tab.Establishments },
+                                new { @class = "govuk-breadcrumbs__link" })
+                        </li>
+                        <li class="govuk-breadcrumbs__list-item">
+                            <a class="govuk-breadcrumbs__link" href=@($"{Url.Action("Index", "EstablishmentsSearch", new { area = "Establishments" })}?{Model.SearchQueryString}")>Search results</a>
+                        </li>
```
RogerHowellDfE, line 42: Note to self -- This is just a whitespace / line break change.

Web/Edubase.Web.UI/Areas/Establishments/Views/EstablishmentsSearch/Downloads/SelectCustomFields.cshtml
```diff
@@ -54,32 +65,33 @@
             <h2 class="govuk-heading-m">Select the data you are interested in</h2>
         </div>
     </div>
-    @using (Html.BeginForm("PrepareDownload", "Search", FormMethod.Get, new Dictionary<string, object> { ["@submit"] = "handleSubmit", ["id"] = "field-selection" }))
+    @using (Html.BeginForm("PrepareDownload", "EstablishmentsSearch", FormMethod.Get,
```
RogerHowellDfE, line 68: Renaming `Search` -> `EstablishmentSearch` appears to be the only non-whitespace change in this file/commit.

There are similar changes renaming `Search` -> `GovernorSearch` / `GroupSearch`.

This parameter appears to be `controllerName`.

@ScottDawsonDfE -- Will changing these controller names have an impact elsewhere? 

I don't see a corresponding rename of files/classes or similar, but I am unfamiliar with this part of the project so not 100% sure what I'm looking for!

Web/Edubase.Services/Nomenclature/NomenclatureService.cs
```diff
@@ -2,6 +2,7 @@
 using Edubase.Common.Text;
 using Edubase.Services.Enums;
 using System;
+using System.Diagnostics;
```
RogerHowellDfE, line 5: Introduction of a new, unused, `using` - a legacy from intermediate refactoring and is to be removed, perhaps?

Web/Edubase.Web.UI/Areas/Establishments/Views/EstablishmentsSearch/Downloads/SelectCustomFields.cshtml
ScottDawsonDfE: There was some weirdness going on before, the methods didn't exist in the controller being pointed to. Thanks to the checking carried out in Rider I was alerted to the wierdness. This is a fix to this.
The changes point the form to point to the correct method in the correct controller.
It is not clear to me how these technically work, but the changes have been manually checked and they seem to be working fine.

Web/Edubase.Services/Nomenclature/NomenclatureService.cs
ScottDawsonDfE: will have a look at this

Web/Edubase.Services.Texuna/ChangeHistory/ChangeHistoryService.cs
ScottDawsonDfE: I will check to see how this is used - removing it could be a better move.

Web/Edubase.Services/Nomenclature/NomenclatureService.cs
```diff
@@ -27,12 +28,25 @@ public string GetGovernorRoleName(eLookupGovernorRole role, eTextCase textCase =
...
+                case GT.ChildrensCentresGroup:
+                    return "children's centres".ToTextCase(textCase);
+                case GT.Federation:
+                case GT.Trust:
+                    return "schools".ToTextCase(textCase);
+                case GT.MultiacademyTrust:
+                case GT.SingleacademyTrust:
+                case GT.SchoolSponsor:
+                    return "academies".ToTextCase(textCase);
+                case eLookupGroupType.UmbrellaTrust:
```
RogerHowellDfE, line 45: For consistency, consider: `GT.UmbrellaTrust` ?

Note that further up the file is this, which makes the `GT` / `eLookupGroupType` effectively interchangeable for now:
```csharp
using GT = eLookupGroupType;
```

Web/Edubase.Services/Nomenclature/NomenclatureService.cs
```diff
@@ -27,12 +28,25 @@ public string GetGovernorRoleName(eLookupGovernorRole role, eTextCase textCase =
...
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(groupType),
+                        $"Group type '{groupType}' is not supported for this operation");
```
RogerHowellDfE, line 48: Using `ArgumentOutOfRangeException` for a non-numeric value jumped out to me as unusual, but can't quite place why. 

Perhaps `NotSupportedException` might match the description more closely?

RogerHowellDfE: Feel free to disregard this observation!

Web/Edubase.Services/Nomenclature/NomenclatureService.cs
```diff
@@ -27,12 +28,25 @@ public string GetGovernorRoleName(eLookupGovernorRole role, eTextCase textCase =
...
+            {
+                case GT.ChildrensCentresCollaboration:
+                case GT.ChildrensCentresGroup:
+                    return "children's centres".ToTextCase(textCase);
+                case GT.Federation:
+                case GT.Trust:
+                    return "schools".ToTextCase(textCase);
+                case GT.MultiacademyTrust:
+                case GT.SingleacademyTrust:
+                case GT.SchoolSponsor:
```
RogerHowellDfE, line 43: It does reduce some readability, but perhaps consider making it explicit where we are deliberately intending to fall-through on the case statements?

Something like this:

```csharp
            switch(groupType)
            {
                case GT.ChildrensCentresCollaboration:
                    // fall-through
                case GT.ChildrensCentresGroup:
                    return "children's centres".ToTextCase(textCase);
                case GT.Federation:
                    // fall-through
                case GT.Trust:
                    return "schools".ToTextCase(textCase);
                case GT.MultiacademyTrust:
                    // fall-through
                case GT.SingleacademyTrust:
                    // fall-through
                case GT.SchoolSponsor:
                    return "academies".ToTextCase(textCase);
                case GT.UmbrellaTrust:
                    // fall-through
                default:
                    throw new ArgumentOutOfRangeException(nameof(groupType),
                        $"Group type '{groupType}' is not supported for this operation");
            }
```

Web/Edubase.Services.Texuna/ChangeHistory/ChangeHistoryService.cs
```diff
@@ -28,13 +28,13 @@ public async Task<List<UserGroupModel>> GetSuggesterGroupsAsync(IPrincipal princ
         [Obsolete("Please use GetSuggesterGroupsAsync", true)]
         public async Task<List<UserGroupModel>> GetApproversGroupsAsync(IPrincipal principal)
         {
-            throw new NotImplementedException("GetApproversGroupsAsync is not implemented because it returns the exact same information as GetSuggesterGroupsAsync");
+            return await GetSuggesterGroupsAsync(principal);
```
RogerHowellDfE, line 31: Calling `GetSuggesterGroupsAsync` within `GetApproversGroupsAsync` looks like a bug, but we can see from the previous implementation this is deliberate.

If we are keeping this change, perhaps we can add a comment indicating the rationale?
```csharp
            // Deliberate - suggesters returns the same as approvers (//todo: elaborate re: if this is referring to the Java API or something else)
            return await GetSuggesterGroupsAsync(principal);
```

Web/Edubase.Services.Texuna/Lookup/LookupApiService.cs
ScottDawsonDfE: Removed commented code

Web/Edubase.Services.TexunaUnitTests/packages.config
```diff
@@ -1,7 +1,9 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Castle.Core" version="5.1.0" targetFramework="net48" />
+  <package id="Microsoft.AspNet.WebApi.Client" version="5.2.9" targetFramework="net48"/>
   <package id="Moq" version="4.18.2" targetFramework="net48" />
+    <package id="Newtonsoft.Json" version="13.0.2" targetFramework="net48"/>
```
RogerHowellDfE, line 6: Minor: inconsistent leading whitespace

Web/Edubase.Services.Texuna/ChangeHistory/ChangeHistoryService.cs
ScottDawsonDfE: It was kept due to it being part of the interface specification - the unused part of the interface has been removed too

Web/Edubase.Services/Nomenclature/NomenclatureService.cs
ScottDawsonDfE: removed the unused using - it was from intermediate steps

Web/Edubase.Services/Nomenclature/NomenclatureService.cs
ScottDawsonDfE: I have changed this - thanks

Web/Edubase.Web.UI/Areas/Establishments/Controllers/BulkCreateFreeSchoolsController.cs
```diff
@@ -128,7 +128,7 @@ private async Task<ActionResult> ResultInternalAsync(Guid id, BulkCreateFreeScho
             }
         }
 
-        private async Task<ActionResult> ResultInternalAjaxAsync(Guid id, BulkCreateFreeSchoolsViewModel viewModel, ApiResponse<BulkCreateFreeSchoolsResult> apiResponse)
+        private ActionResult ResultInternalAjaxAsync(Guid id, BulkCreateFreeSchoolsViewModel viewModel, ApiResponse<BulkCreateFreeSchoolsResult> apiResponse)
```
RogerHowellDfE, line 131: I really struggled to figure out the original nested ternarys to figure out if the change is equivalent (I assume it is, just trying to rationalise it and be confident in my own mind).

If we're refactoring this piece of code perhaps we would consider removing the nested ternary operators in favour of simpler if statements? 

Personally I find the example below easier to reason about / follow at-a-glance, but appreciate it's somewhat a stylistic view.

_The below was produced using IDE refactorings ("convert to if" and "invert if") so I don't have any reason to doubt this being equivalent, but worth double checking for yourself before using as-is._

```csharp
        private ActionResult ResultInternalAjaxAsync(Guid id, BulkCreateFreeSchoolsViewModel viewModel, ApiResponse<BulkCreateFreeSchoolsResult> apiResponse)
        {
            var redirectUrl = string.Concat("/Establishments/bulk-create-free-schools/", id);
            if (!apiResponse.Success)
            {
                if (!apiResponse.HasErrors)
                {
                    throw new Exception("ApiResponse indicated failure, but no errors were supplied");
                }

                return Json(JsonConvert.SerializeObject(new {status = true, redirect = redirectUrl}));
            }

            viewModel.Result = apiResponse.GetResponse();
            if (viewModel.Result.IsProgressing())
            {
                return Json(JsonConvert.SerializeObject(new {status = false, redirect = redirectUrl}));
            }

            if (viewModel.Result.IsCompleted())
            {
                return Json(JsonConvert.SerializeObject(new {status = true, redirect = redirectUrl}));
            }

            throw new Exception($"The status of task {id} is unclear; the API did not provide a good enough response {Newtonsoft.Json.JsonConvert.SerializeObject(viewModel.Result)}");

        }
```

Web/Edubase.Web.UIUnitTests/Areas/Establishments/Controllers/EstablishmentControllerTests.cs
```diff
@@ -313,8 +313,7 @@ public async Task Estab_AddOrReplaceEstablishmentAddress()
             var result = await controller.AddOrReplaceEstablishmentAddressAsync(5, "test");
 
             Assert.NotNull(result);
-            //TODO - The above assert was added when converted to XUnit, before this the test asserted nothing.
-            //Figure out if anything else needs testing on this, or does it it just test the method completes without exception
```
RogerHowellDfE, line 317: @ScottDawsonDfE -- Can we assume you've checked and this TODO no longer applies? :)

Web/Edubase.Services/Nomenclature/NomenclatureService.cs
ScottDawsonDfE: Happy to have discussion on this, as it may help broaden understanding:
I use ArgumentOutOfRangeException for something like this (taken from https://www.pluralsight.com/guides/commonly-encountered-exceptions-ArgumentOutOfRangeException)

> The ArgumentOutOfRangeException exception is thrown when the argument passed to a method is not null and contains a value that is not within the set of expected values. This exception type has the properties [ParamName](https://docs.microsoft.com/en-us/dotnet/api/system.argumentexception.paramname?view=netframework-4.7.2) and [ActualValue](https://docs.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception.actualvalue?view=netframework-4.7.2) which helps understand the cause of the exception. The ParamName property identifies the parameter name of the invalid argument and the ActualValue property identifies the invalid value if a value is present.

I have added a bit onto the use of the exception to provide the value that is causing issues.

When the issue relates to a parameter passed to a method, I do prefer to use argument exceptions too, as they are written for raising iussues with passed parameters.

Web/Edubase.Services/Nomenclature/NomenclatureService.cs
ScottDawsonDfE: Interesting point.

I think this is down to how I read case statements and the model that exists in my head for them. I prefer without the comments, I would rather copy and paste the return lines, than the commnets, but also don't like duplicated code like this.

Happy to go through this

Web/Edubase.Services.Texuna/ChangeHistory/ChangeHistoryService.cs
ScottDawsonDfE: This should now be removed

Web/Edubase.Services.TexunaUnitTests/packages.config
ScottDawsonDfE: resolved

Web/Edubase.Web.UIUnitTests/Areas/Establishments/Controllers/EstablishmentControllerTests.cs
ScottDawsonDfE: ;-) Lets assume there is nothing todo

There are improvements that can be made to the tests, but for replacing the existing tests (work I did this time last year), this is sufficient

Web/Edubase.Web.UI/Areas/Establishments/Controllers/BulkCreateFreeSchoolsController.cs
ScottDawsonDfE: There are two changes here.
The method signature - which removes the async and task part, as nothing async happens in the method

Then the if statement. As this is resolving IDE warnings and problems, I haven't widened the scope.
The bit changed gives the same result whether the result "HasCreatedEstablishments" or not - the JSON produced for both was exactly the same.

However much I would like to widen the scope of these changes, I want to keep a focus on only resolving IDE highlighted issues - otherwise I could be doing this for weeks.

Web/Edubase.Services/Nomenclature/NomenclatureService.cs
RogerHowellDfE: I think it comes down to it "feeling" strange to consider an enum as having a "range". 

If implementing fresh I think I would have defaulted to using either `NotSupportedException` re: the implementation not yet existing, or `ArgumentException`/`InvalidEnumArgumentException` if it's not an appropriate value passed in.

Not that it particularly matters as in this case it's all a bit academic beyond flagging there's an error here 😀 

_(I don't have rights to mark this comment as resolved - feel free to do it on my behalf here!)_

Web/Edubase.Services/Nomenclature/NomenclatureService.cs
RogerHowellDfE: Feel free to resolve/disregard this suggestion and leave as-is :)

Web/Edubase.Services/Nomenclature/NomenclatureService.cs
ScottDawsonDfE: Changed to InvalidEnumArgumentException

RogerHowellDfE (approved):

RogerHowellDfE:
I see one remaining warning:

image


```
Build started...
1>------ Build started: Project: Edubase.Common, Configuration: Debug Any CPU ------
Restored C:\dev\get-information-about-schools\Web\Edubase.CommonUnitTests\Edubase.CommonUnitTests.csproj (in 11 ms).
Restored C:\dev\get-information-about-schools\Web\Edubase.Web.ResourcesUnitTests\Edubase.Web.ResourcesUnitTests.csproj (in 11 ms).
Restored C:\dev\get-information-about-schools\Web\Edubase.ServicesUnitTests\Edubase.ServicesUnitTests.csproj (in 12 ms).
Restored C:\dev\get-information-about-schools\Web\Edubase.Web.UIUnitTests\Edubase.Web.UIUnitTests.csproj (in 14 ms).
Restored C:\dev\get-information-about-schools\Web\Edubase.DataUnitTests\Edubase.DataUnitTests.csproj (in 20 ms).
1>  Edubase.Common -> C:\dev\get-information-about-schools\Web\Edubase.Common\bin\Debug\Edubase.Common.dll
2>------ Build started: Project: Edubase.CommonUnitTests, Configuration: Debug Any CPU ------
3>------ Build started: Project: Edubase.Data, Configuration: Debug Any CPU ------
2>  Edubase.CommonUnitTests -> C:\dev\get-information-about-schools\Web\Edubase.CommonUnitTests\bin\Debug\Edubase.CommonUnitTests.dll
3>  Edubase.Data -> C:\dev\get-information-about-schools\Web\Edubase.Data\bin\Debug\Edubase.Data.dll
4>------ Build started: Project: Edubase.DataUnitTests, Configuration: Debug Any CPU ------
5>------ Build started: Project: Edubase.Services, Configuration: Debug Any CPU ------
4>  Edubase.DataUnitTests -> C:\dev\get-information-about-schools\Web\Edubase.DataUnitTests\bin\Debug\Edubase.DataUnitTests.dll
5>  Edubase.Services -> C:\dev\get-information-about-schools\Web\Edubase.Services\bin\Debug\Edubase.Services.dll
6>------ Build started: Project: Edubase.ServicesUnitTests, Configuration: Debug Any CPU ------
7>------ Build started: Project: Edubase.Web.Resources, Configuration: Debug Any CPU ------
8>------ Build started: Project: Edubase.Services.Texuna, Configuration: Debug Any CPU ------
7>C:\dev\get-information-about-schools\Web\packages\Microsoft.Bcl.Build.1.0.21\build\Microsoft.Bcl.Build.targets(243,5): warning : All projects referencing Edubase.Services.csproj must install nuget package Microsoft.Bcl.Build. For more information, see http://go.microsoft.com/fwlink/?LinkID=317569.
7>  Edubase.Web.Resources -> C:\dev\get-information-about-schools\Web\Edubase.Web.Resources\bin\Debug\Edubase.Web.Resources.dll
9>------ Build started: Project: Edubase.Web.ResourcesUnitTests, Configuration: Debug Any CPU ------
9>C:\dev\get-information-about-schools\Web\packages\Microsoft.Bcl.Build.1.0.21\build\Microsoft.Bcl.Build.targets(243,5): warning : All projects referencing Edubase.Services.csproj must install nuget package Microsoft.Bcl.Build. For more information, see http://go.microsoft.com/fwlink/?LinkID=317569.
9>Done building project "Edubase.Services.csproj".
6>  Edubase.ServicesUnitTests -> C:\dev\get-information-about-schools\Web\Edubase.ServicesUnitTests\bin\Debug\Edubase.ServicesUnitTests.dll
8>  Edubase.Services.Texuna -> C:\dev\get-information-about-schools\Web\Edubase.Services.Texuna\bin\Debug\Edubase.Services.Texuna.dll
10>------ Build started: Project: Edubase.Web.UI, Configuration: Debug Any CPU ------
10>  Edubase.Web.UI -> C:\dev\get-information-about-schools\Web\Edubase.Web.UI\bin\Edubase.Web.UI.dll
10>  Microsoft.ApplicationInsights.Web: Copying ApplicationInsights.config to bin\
11>------ Build started: Project: Edubase.Services.TexunaUnitTests, Configuration: Debug Any CPU ------
12>------ Build started: Project: Edubase.Web.UIUnitTests, Configuration: Debug Any CPU ------
9>Edubase.Web.ResourcesUnitTests -> C:\dev\get-information-about-schools\Web\Edubase.Web.ResourcesUnitTests\bin\Debug\net48\Edubase.Web.ResourcesUnitTests.dll
11>  Edubase.Services.TexunaUnitTests -> C:\dev\get-information-about-schools\Web\Edubase.Services.TexunaUnitTests\bin\Debug\Edubase.Services.TexunaUnitTests.dll
12>Edubase.Web.UIUnitTests -> C:\dev\get-information-about-schools\Web\Edubase.Web.UIUnitTests\bin\Debug\net48\Edubase.Web.UIUnitTests.dll
========== Build: 12 succeeded, 0 failed, 0 up-to-date, 0 skipped ==========
========== Elapsed 00:05.520 ==========
```

ScottDawsonDfE:
Yep,

Fixing that one properly is going to require moving to .net standard. Trying to add the dependency does nothing (or I’m doing it wrong).

My proper fix is going to be the complete removal of BCL.Build

Thanks,
Scott

anthony9853 (approved):

PR #268, opened 2023-01-17
added id locators for children's centre automated tests

antk43:


Web/Edubase.Web.UI/Areas/Establishments/Views/Establishment/Details.cshtml
```diff
@@ -71,7 +71,7 @@
                 <p class="message-text" aria-live="polite">
                     @if (approvedCount > 0)
                     {
-                        <div class="tick"></div>
+                        <div id="update-banner-tick-icon" class="tick"></div>
```
RogerHowellDfE, line 74: Is there a circumstance where this `div` and the `div` on line 53 will be shown at the same time? I do not know the page/logic well enough to be sure.

Given that they have the same ID value, there is potential for two HTML elements with the same ID may be on the same page and accessing the element by ID will vary / will potentially be inconsistent.

This may also lead to invalid test results depending on how they're being used (e.g., expecting element with this ID to be present, and it is present but not for the expected reason/combination of values).

antk43: Great spot. There probably is an instance so let me change it.

PR #270, opened 2023-01-20
Added ids to continue button and  download XLSX button

JaswiniP:
added ids for continue button and download XLSX button

### DIFF
--- a/Web/Edubase.Web.UI/Views/Tools/IndSchoolSearchDownload/ReadyToDownload.cshtml
+++ b/Web/Edubase.Web.UI/Views/Tools/IndSchoolSearchDownload/ReadyToDownload.cshtml
@@ -53,7 +53,7 @@
                 <p>Your requested file is ready to download.</p>
 
                 <p>
-                    <input type="submit" class="govuk-button" data-module="govuk-button" value="@name" /><br />
+                    <input type="submit" id="@StringUtils.ElementIdFormat(@name)-button" class="govuk-button" data-module="govuk-button" value="@name" /><br />
                     (will download as a ZIP file: open the ZIP to access its content)
                 </p>
             </div>
--- a/Web/Edubase.Web.UI/Views/Tools/IndependentSchoolsSearch.cshtml
+++ b/Web/Edubase.Web.UI/Views/Tools/IndependentSchoolsSearch.cshtml
@@ -106,7 +106,7 @@
 
                 <div class="govuk-grid-column-full">
                     <div class="button-row govuk-!-padding-top-8">
-                        <button type="submit" name="@VM.BindAliasForActionName" class="govuk-button" value="@VM.ActionSearch">Continue</button>
+                        <button type="submit" id="continue-button" name="@VM.BindAliasForActionName" class="govuk-button" value="@VM.ActionSearch">Continue</button>
                         @Html.ActionLink("Cancel", "Index", "Tools", null, new {@class = "govuk-button govuk-button--secondary", data_module = "govuk-button"})
                     </div>
                 </div>


Web/Edubase.Web.UI/Views/Tools/IndSchoolSearchDownload/ReadyToDownload.cshtml
```diff
@@ -53,7 +53,7 @@
                 <p>Your requested file is ready to download.</p>
 
                 <p>
-                    <input type="submit" class="govuk-button" data-module="govuk-button" value="@name" /><br />
+                    <input type="submit" id="@StringUtils.ElementIdFormat(@name)-button" class="govuk-button" data-module="govuk-button" value="@name" /><br />
```
CathLass, line 56: I don't have the knowledge to know that the variable 'name' will always contain a string that can be successfully formatted to produce a valid id here - but if you know that I'm happy.

CathLass (approved):

Web/Edubase.Web.UI/Views/Tools/IndSchoolSearchDownload/ReadyToDownload.cshtml
JaswiniP: the @name is referred to the name of the download

antk43: Thanks @CathLass yes the name value will always contain one of two strings which can be formatted successfully.

RogerHowellDfE, line 56: It would be worth double checking if this implementation is sufficient to meet this need, and would be robust enough. 

My thought process (and current assumption) is that `@name` refers to a file name or perhaps a display name?

If so, how does it handle apostrophes like for Children's Centres:

image

```csharp
// Current implementation:

namespace Edubase.Web.UI.Helpers
{
    public class StringUtils
    {
        public static string ElementIdFormat(string text)
        {
            return text.Replace(' ', '-').ToLowerInvariant();
        }
    }
}
```

```csharp
// Suggested implementation

using System;
using System.Text.RegularExpressions;
using Edubase.Web.UI.Helpers;


public class Program
{
    public static void Main()
    {
        Console.WriteLine("Hello World");
        Console.WriteLine(StringUtils.ElementIdFormat("example with spaces"));
        Console.WriteLine(StringUtils.ElementIdFormat("example with ' apostrophes"));
        Console.WriteLine(StringUtils.ElementIdFormat("example with \" quotes"));
        Console.WriteLine(StringUtils.ElementIdFormat("example with [] square brackets"));
        Console.WriteLine(StringUtils.ElementIdFormat("example with () round brackets"));
        Console.WriteLine(StringUtils.ElementIdFormat("example with accents -- The most common accents are the acute (é), grave (è), circumflex (â, î or ô), tilde (ñ), umlaut and dieresis (ü or ï – the same symbol is used for two different purposes), and cedilla (ç)."));
    }
}


namespace Edubase.Web.UI.Helpers
{
    public class StringUtils
    {
        public static string ElementIdFormat(string text)
        {
            var pattern = @"[^a-zA-Z0-9_]";
            var replacement = "-";
            var result = Regex.Replace(text, pattern, replacement);

            return result.ToLowerInvariant();
        }
    }
}
```


https://dotnetfiddle.net/X4b31o

image

antk43: Thanks @RogerHowellDfE this is great. I've made some updates to the StringUtils class in my latest PR but I will look to implement these changes as well.

RogerHowellDfE: Ah, looks like @CathLass beat me to this comment (I was too busy tinkering 😁 ) and this has already been explained by @JaswiniP and resolved as a non-issue before I hit submit and refreshed the page -- I withdraw my comment!

For further info/context, it seems that `name` refers to the file type name from a constrained list of values, which should never have unusual characters so this PR is fine for this specific page 👍 

image

RogerHowellDfE: For further info/context using a screenshot from @JaswiniP (thanks!) it seems that `name` refers to the file type name from a constrained list of values, which should never have unusual characters 👍 

image

PR #271, opened 2023-01-20
id locators added for childrens centre group pages

antk43:


Web/Edubase.Web.UI/Areas/Groups/Views/Group/CreateChildrensCentre.cshtml
```diff
@@ -31,37 +31,37 @@
...
-                        <a href="#LinkedEstablishments_LinkedEstablishmentSearch_Urn">The <abbr title="Unique Reference Number">URN</abbr> is a duplicate</a>
+                        <a id="duplicate-urn-error" href="#LinkedEstablishments_LinkedEstablishmentSearch_Urn">The <abbr title="Unique Reference Number">URN</abbr> is a duplicate</a>
                     </li>
                     <li v-show="apiErrors.length>0" v-for="err in apiErrors">
-                        <a href="#LinkedEstablishments_LinkedEstablishmentSearch_Urn">{{err.field}} <span v-if="err.field">:</span> {{err.message}}</a>
+                        <a id="api-error" href="#LinkedEstablishments_LinkedEstablishmentSearch_Urn">{{err.field}} <span v-if="err.field">:</span> {{err.message}}</a>
                     </li>
                     <li v-show="joinDateError">
-                        <a href="#JoinedDate">Please check the joined date you have entered</a>
+                        <a id="joine-date-error" href="#JoinedDate">Please check the joined date you have entered</a>
```
CathLass, line 64: typo? joine-data

CathLass (approved):

Web/Edubase.Web.UI/Areas/Groups/Views/Group/EditLinks.cshtml
```diff
@@ -128,7 +128,7 @@
                     <div class="button-row govuk-!-margin-bottom-4">
                         @if (!Model.LinkedEstablishments.Establishments.Any(x => x.EditMode))
                         {
-                            <button type="submit" name="action" class="govuk-button" value="@VM.ActionSaveLinks">Save @Model.ListOfEstablishmentsPluralName.ToLower()</button>
+                            <button id="save-@StringUtils.ElementIdFormat(@Model.ListOfEstablishmentsPluralName)-top-button" type="submit" name="action" class="govuk-button" value="@VM.ActionSaveLinks">Save @Model.ListOfEstablishmentsPluralName.ToLower()</button>
```
RogerHowellDfE, line 131: A copy/pasted of my comment on the other PR which looks like it might apply here if it's using the establishment name.

Note that establishment names can contain additional characters which are invalid for HTML ID values -- URN 155806 on our test environment is a particularly "special" example.




----


It would be worth double checking if this implementation is sufficient to meet this need, and would be robust enough. 

My thought process (and current assumption) is that `@name` refers to a file name or perhaps a display name?

If so, how does it handle apostrophes like for Children's Centres (noting that apostrophes are not valid characters for an ID value):

image

```csharp
// Current implementation:

namespace Edubase.Web.UI.Helpers
{
    public class StringUtils
    {
        public static string ElementIdFormat(string text)
        {
            return text.Replace(' ', '-').ToLowerInvariant();
        }
    }
}
```

```csharp
// Suggested implementation

using System;
using System.Text.RegularExpressions;
using Edubase.Web.UI.Helpers;


public class Program
{
    public static void Main()
    {
        Console.WriteLine("Hello World");
        Console.WriteLine(StringUtils.ElementIdFormat("example with spaces"));
        Console.WriteLine(StringUtils.ElementIdFormat("example with ' apostrophes"));
        Console.WriteLine(StringUtils.ElementIdFormat("example with \" quotes"));
        Console.WriteLine(StringUtils.ElementIdFormat("example with [] square brackets"));
        Console.WriteLine(StringUtils.ElementIdFormat("example with () round brackets"));
        Console.WriteLine(StringUtils.ElementIdFormat("example with accents -- The most common accents are the acute (é), grave (è), circumflex (â, î or ô), tilde (ñ), umlaut and dieresis (ü or ï – the same symbol is used for two different purposes), and cedilla (ç)."));
    }
}


namespace Edubase.Web.UI.Helpers
{
    public class StringUtils
    {
        public static string ElementIdFormat(string text)
        {
            var pattern = @"[^a-zA-Z0-9_]";
            var replacement = "-";
            var result = Regex.Replace(text, pattern, replacement);

            return result.ToLowerInvariant();
        }
    }
}
```


https://dotnetfiddle.net/X4b31o

image

Web/Edubase.Web.UI/Areas/Groups/Views/Group/CreateChildrensCentre.cshtml
antk43: Great spot. Typo now fixed.

Web/Edubase.Web.UI/Areas/Groups/Views/Group/EditLinks.cshtml
antk43: Added your changes in with the extra string.empty replacements.

RogerHowellDfE (approved):

PR #274, opened 2023-01-27
Add omitted governance professional to enumset

RogerHowellDfE:
The collection now matches the database and the listed description.

Proposed changes/edits:
- All six governance professional types listed
- Test cases validate this/prevent future accidental regressions

#140851

### DIFF
--- a/Web/Edubase.Services/Enums/EnumSets.cs
+++ b/Web/Edubase.Services/Enums/EnumSets.cs
@@ -39,6 +39,7 @@ namespace Edubase.Services.Enums
             ET.UniversityTechnicalCollege
         }.Cast<int>();
 
+        // See also database table StaffRole, column isOnePersonRole
         public static IEnumerable<GR> eSingularGovernorRoles { get; } = new[]
         {
             GR.ChairOfGovernors,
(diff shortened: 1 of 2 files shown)


Web/Edubase.Services/Enums/EnumSets.cs
```diff
@@ -51,11 +52,13 @@ public static class EnumSets
             GR.GovernanceProfessionalToAFederation,
             GR.GovernanceProfessionalToAnIndividualAcademyOrFreeSchool,
             GR.GovernanceProfessionalToAMat,
+            GR.Group_SharedGovernanceProfessional,
```
ScottDawsonDfE, line 55: Will need to check hoiw this actually end up working

The Group_SharedGovernanceProfessional role, is one that would be applied at a MAT, and theoretically the MAT could have a pool of shared professiionals, with each establishment only using one of them - through the Establishment_SharedGovernanceProfessional role.

At the moment, I am not sure we want to add this role to the singular roles

ScottDawsonDfE (approved):

Web/Edubase.Services/Enums/EnumSets.cs
RogerHowellDfE: Verbal discussion - decision: include it now, but review it later when looking specifically at the group implementations.

RogerHowellDfE:
Comments added to future work items to highlight this discussion outcome.

Merging.

PR #275, opened 2023-01-27
Restrict creating new if would overlap with current governance professional

RogerHowellDfE:
## Background

- There is pre-existing functionality to prevent creating a new chair of a local governing body where it (or the shared equivalent) already exists

## Proposed changes

- Extend pre-existing functionality/behaviour to also apply validation rules permitting only a single appointment of any type of governance professional

**Creating as draft PR for now as unclear if this is to be implemented within frontend/backend(/both).**

### DIFF
--- a/Web/Edubase.Services/Enums/EnumSets.cs
+++ b/Web/Edubase.Services/Enums/EnumSets.cs
@@ -82,6 +82,15 @@ namespace Edubase.Services.Enums
             GR.Establishment_SharedGovernanceProfessional,
         };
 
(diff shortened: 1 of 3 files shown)


RogerHowellDfE:
Note, this is how it appears/displays on screen when attempting to appoint a new chairperson (and now, with this PR, governance professionals) where one is currently in post and active.

Note that this restricts access to the "add new" form, and is separate to validations performed when submitting the form (presumably preferred, to prevent a user spending time filling out the form only to find out it would _never_ be accepted regardless of what is entered into the form).

image

ScottDawsonDfE (approved):

PR #280, opened 2023-02-08
new values using 46 for SecureAcademies16to19 type

anthony9853:


ScottDawsonDfE:
Check which branch you want to merge into

Web/Edubase.Services.Texuna/Establishments/EstablishmentReadApiService.cs
```diff
@@ -156,7 +156,9 @@ public async Task<EstablishmentEditPolicyEnvelope> GetEditPolicyByUrnAsync(int u
                 { ET.ChildrensCentreLinkedSite, new[] { EP.NotApplicable } },
                 { ET.InstitutionFundedByOtherGovernmentDepartment, new[] { EP.NotApplicable } },
```
ScottDawsonDfE, line 158: The empty lines in this are a little weird, I get that you are following exisinting convention here. But we should be avoiding unneccassary whitespace. It may be worth removing the other empty lines in this section that aren't needed.

ScottDawsonDfE (changes requested):
Don't merge to master
Not sure if you want this to dev, or to an epic branch

Web/Edubase.Services.Texuna/Establishments/EstablishmentReadApiService.cs
anthony9853: assuming this is ok as is

PR #291, opened 2023-02-17
Updated bulk update guidance to use just LA 

ScottDawsonDfE:
- renaming LA (Name), and removing LA Code. Also applied lint/reformat of the file

### DIFF
--- a/Web/Edubase.Web.UI/Views/Guidance/EstablishmentBulkUpdate.cshtml
+++ b/Web/Edubase.Web.UI/Views/Guidance/EstablishmentBulkUpdate.cshtml
@@ -32,840 +32,837 @@
 </div>
 
 <div class="govuk-grid-row">
-    <div class="govuk-grid-column-full">
-        <h2 class="govuk-heading-m">Introduction</h2>
-        <p>The bulk update tool allows you to make many updates in a single batch instead of editing individual establishment records. Multiple column files must be XLSX and bulk upload files can be XLSX or tab delimited text files.</p>
-        <p>If there&rsquo;s an error in your data, we&rsquo;ll send you a message to correct it. If your changes need approval, you&rsquo;ll get a second message to say your change is either approved or declined.</p>
-        <p>If there&rsquo;s an error in your data, we&rsquo;ll send you a message to correct it.</p>
-        <p>If your changes need approval, you&rsquo;ll get a second message to say your change is either approved or declined.</p>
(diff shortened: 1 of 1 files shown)


ScottDawsonDfE:
Note: most of the cahnges are fixing the formatting of the file. Unfortunately it makes spotting the two bits of real code change difficult
Line 189 in the new file
And lines 251-254 removed from the old file

RogerHowellDfE:
> Note: most of the changes are fixing the formatting of the file. Unfortunately it makes spotting the two bits of real code change difficult Line 189 in the new file And lines 251-254 removed from the old file

To make this easier to review, could you split this into two separate commits - one to do the change, and a second commit to do the reformat? (or the other way around)

Web/Edubase.Web.UI/Views/Guidance/EstablishmentBulkUpdate.cshtml
```diff
@@ -32,840 +32,837 @@
...
+    <td class="govuk-table__cell" data-label="Equivalent field name in app">County (name)</td>
+</tr>
+<tr class="govuk-table__row">
+    <td class="govuk-table__cell" data-label="Database name (use this for your upload file)">Postcode</td>
+    <td class="govuk-table__cell" data-label="Equivalent field name in app">Postcode</td>
+</tr>
+<tr class="govuk-table__row">
+    <td class="govuk-table__cell" data-label="Database name (use this for your upload file)">LA</td>
+    <td class="govuk-table__cell" data-label="Equivalent field name in app">Local authority</td>
+</tr>
```
RogerHowellDfE, line 191: It is now ambiguous which value should be provided (name or ID or code).

Reading through the bug comments I _think_ it's the LA code which is expected.

Either way, elaborating here would be helpful I think.

RogerHowellDfE (changes requested):

BarryGibney (approved):

Web/Edubase.Web.UI/Views/Guidance/EstablishmentBulkUpdate.cshtml
ScottDawsonDfE: Discussed with @RogerHowellDfE 
This is ambiguous as the value updates the LA using either Code or Name.
There are much bigger questions about the guidance as updating LA also updates other fields

RogerHowellDfE (approved):

PR #300, opened 2023-02-22
Additional ID locators to match the test suite requirements

lpotter-dfe:
Another locator train for you all. I thought I had added these ones in but apparently not
Again a single file with a few ID's added to make life easier for the test suite

### DIFF
--- a/Web/Edubase.Web.UI/Views/Shared/EditorTemplates/GovernorViewModel.cshtml
+++ b/Web/Edubase.Web.UI/Views/Shared/EditorTemplates/GovernorViewModel.cshtml
@@ -81,19 +81,19 @@ else
         <div class="govuk-form-group @Html.ValidationCssClassFor(x => x.FirstName)">
             @Html.LabelFor(x => x.FirstName, $"First name{additionalInfo}", new { @class = "govuk-label" })
             @Html.ValidationMessageFor(x => x.FirstName, null, new { @class = "govuk-error-message" })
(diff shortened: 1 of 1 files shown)


RogerHowellDfE (approved):

Web/Edubase.Web.UI/Views/Shared/EditorTemplates/GovernorViewModel.cshtml
```diff
@@ -81,19 +81,19 @@ else
         <div class="govuk-form-group @Html.ValidationCssClassFor(x => x.FirstName)">
             @Html.LabelFor(x => x.FirstName, $"First name{additionalInfo}", new { @class = "govuk-label" })
             @Html.ValidationMessageFor(x => x.FirstName, null, new { @class = "govuk-error-message" })
-            @Html.TextBoxFor(x => x.FirstName, new { @class = string.Concat("govuk-input ", Html.TextBoxValidationClass(x => x.FirstName)) })
+            @Html.TextBoxFor(x => x.FirstName, new { @class = string.Concat("govuk-input ", Html.TextBoxValidationClass(x => x.FirstName)), id = "governorFirstNameInput" })
```
antk43, line 84: We've been using the naming convention of dashes between words and lower casing so could you change all the ids here to that i.e. "governor-first-name-input".
Also, on the note of the use of 'input' I have been using 'textbox' when the element is as such - thoughts?

lpotter-dfe: That should be in line with our naming convention now, let me know what you think @antk43

antk43 (approved):

PR #301, opened 2023-02-22
Academic sponsor group

JaswiniP:
update couple of locators

### DIFF
--- a/Web/Edubase.Web.UI/Assets/Scripts/Entry/merge-amalgamate.js
+++ b/Web/Edubase.Web.UI/Assets/Scripts/Entry/merge-amalgamate.js
@@ -425,7 +425,7 @@ const mergersApp = new Vue({
 
       if (this.mergeDateError) {
         this.errors.push({
-          href: '#amalgamationdate-day',
+          href: '#amalgamation-date-day',
           message: 'You have entered an invalid date. Please correct the date to continue',
         });
       }
--- a/Web/Edubase.Web.UI/Views/Tools/MergersTool.cshtml
+++ b/Web/Edubase.Web.UI/Views/Tools/MergersTool.cshtml
@@ -260,8 +260,8 @@
                         <div class="govuk-date-input">
                             <div class="govuk-date-input__item">
                                 <div class="govuk-form-group">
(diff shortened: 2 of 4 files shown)


Web/Edubase.Web.UI/Assets/Scripts/Entry/merge-amalgamate.js
```diff
@@ -425,7 +425,7 @@ const mergersApp = new Vue({
 
       if (this.mergeDateError) {
         this.errors.push({
-          href: '#amalgamationdate-day',
+          href: '#amalgamation-date-day',
```
antk43, line 428: @JaswiniP should the value be changed here as we are only changing ids? FYI @RogerHowellDfE

Web/Edubase.Web.UI/Views/Tools/MergersTool.cshtml
```diff
@@ -260,8 +260,8 @@
                         <div class="govuk-date-input">
                             <div class="govuk-date-input__item">
                                 <div class="govuk-form-group">
-                                    <label for="amalgamationdate-day" class="govuk-label govuk-date-input__label">Day</label>
-                                    <input class="govuk-input date-text-day" type="text" v-model="mergeDateDay" id="amalgamationdate-day">
+                                    <label for="amalgamation-date-day" class="govuk-label govuk-date-input__label">Day</label>
```
antk43, line 263: @JaswiniP should the for be changed here as we are only changing ids? FYI @RogerHowellDfE

RogerHowellDfE: The `for` attribute of a `<label>` needs to match the `id` value of the `<input>` element the label is describing - updating one needs the other to be updated also, so updating both here is fine :)

Web/Edubase.Web.UI/Assets/Scripts/Entry/merge-amalgamate.js
RogerHowellDfE: An `href` with an `#` in it is (usually) a link to the HTML element with that `id` on the page (similar to CSS selectors). 

If updating the ID, it makes sense to also update links which point to it so looks good here :)

PR #306, opened 2023-02-22
Bg/144969 bug appointed by added

BarryGibney:


Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
```diff
@@ -159,9 +159,16 @@ public GovernorsGridViewModel()
 
                     if (EnumSets.eGovernanceProfessionalRoles.Contains(role))
                     {
+                        var displayAppointingBody = false;
+                        if (displayPolicy.AppointingBodyId == true && !GovernorPermissions.Update == true)
```
RogerHowellDfE, line 163: Could this be simplified?

```cs
!GovernorPermissions.Update == true
// vs
GovernorPermissions.Update == false
```

RogerHowellDfE: I'm also not quite sure of the intention of that part of the condition - "only display the appointing body if the user has permission to view it, but not permission to update/edit the governor"?

It sounds strange to me, but perhaps there's something I'm not understanding?

BarryGibney: I might not have done it the best way!
                grid.AddHeaderCell("Name", displayPolicy.FullName, "name", "sortText")
                    .AddHeaderCell("Governance role identifier (GID)", displayPolicy.Id, "gid")
                    .AddHeaderCell("Appointed By", displayAppointingBody, "sortText")

I added the .AddHeaderCell ( and in here I wanted to use the displayPolicy but it is always true)
So I created a variable for displayAppointingBody, then it can still do the display policy check (in cases of future changes) along with a conditional check (uses the update to establish if a user is logged in or out) and passes it into the grid.

ScottDawsonDfE: Looking at the 3 lines of code, I would suggest the following, we are assigning a bool and switching on bool values - so can avoid the if statemnet entirely

var displayAppointingBody = displayPolicy.AppointingBody && !GovernorPermissions.Update

Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
```diff
@@ -225,8 +232,14 @@ public GovernorsGridViewModel()
         {
             if (EnumSets.eGovernanceProfessionalRoles.Contains(role))
             {
+                var displayAppointingBody = false;
```
ScottDawsonDfE, line 235: This boolean value and if block can be simplified too

Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
ScottDawsonDfE: For @RogerHowellDfE 
The reason its not displayed when viewing as an updater, is to match the UED (who see the items in these gridviews differently)
The logged in view (i.e. somebody who can update) is to not show the appointing body. Public users (non-editors) should have the appointing body in the view they get.

RogerHowellDfE: @ScottDawsonDfE - Are "logged in"/"somebody who can update" the same thing? Ditto "public users"/"non-editors"? If not, we need to select a different property to use in the boolean.

Notes:
- "public user" has a very specific definition based on roles/user groups the current user has, and
- some users can be logged in but not be able to update a given record - in this case they'll be denied seeing the appointing body (something which a public / not-logged-in user _can_ see?

RogerHowellDfE: More fundamentally, it seems strange that somebody who is not logged in can see more information than somebody who is logged in - if we are simply trying to match the UED, are we sure that this isn't a UED problem here?

ScottDawsonDfE:
Closing pull request as change is not needed

PR #340, opened 2023-03-22
Changed the configuration of the lookup client to be more explicit on the config

ScottDawsonDfE:
Changed the configuration of the lookup client to be more explicit the config - added unit tests for method

### DIFF
--- a/Web/Edubase.Web.UIUnitTests/App_Start/IocConfigTests.cs
+++ b/Web/Edubase.Web.UIUnitTests/App_Start/IocConfigTests.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.Net.Http.Headers;
+using Edubase.Services.Texuna.Core;
+using Edubase.Web.UI;
+using Xunit;
+
+namespace Edubase.Web.UIUnitTests
+{
+    public class IocConfigTests
+    {
+        [Theory]
+        [InlineData("", null, null)]
+        [InlineData(" ", null, null)]
+        [InlineData("     ", null, null)]
+        [InlineData("test", null, null)]
+        public void CreateLookupClient_ThrowsUriFormatExceptionForInvalidAddresses(string lookupApiAddress,
+            string lookupApiUsername, string lookupApiPassword)
+        {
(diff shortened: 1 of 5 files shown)


Web/Edubase.Web.UIUnitTests/App_Start/IocConfigTests.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            Assert.Throws<ArgumentNullException>(() => IocConfig.CreateLookupClient(
+                null,
+                null,
+                null));
+        }
+
+        [Theory]
+        [InlineData(@"https://google.com/")]
+        [InlineData(@"https://duckduckgo.com/")]
```
RogerHowellDfE, line 36: Unless there's a specific reason not to, let's tweak these URLs to `https://example.com/` and `https://example.org/` domains rather than "real" addresses 👍 


Note that there is a [known issue/requirement](https://stackoverflow.com/a/23438417) where the base address _**must**_ have a trailing slash else the trailing segment gets dropped and the URL gets corrupted.

Given this, I also suggest we include test cases where the base address contains a sub-directory component with/without trailing slash (`/`) character, though I have no strong feeling re: if this is done here versus in a separate PR.

Example:

```csharp
        [Theory]
        [InlineData(@"https://example.com/")]
        [InlineData(@"https://example.org/")]
        [InlineData(@"https://example.org/api")]      // shouldn't be permitted
        [InlineData(@"https://example.org/api/")]
        [InlineData(@"https://example.org/api/v1")]   // shouldn't be permitted
        [InlineData(@"https://example.org/api/v1/")]
```

anthony9853 (approved):

Web/Edubase.Web.UIUnitTests/App_Start/IocConfigTests.cs
CathLass: I've acted on the first part of your comment.
The issue around the requirement to have a trailing '/' I'll leave as a separate piece of work.

PR #344, opened 2023-03-23
Removed quiet handling of error...

ScottDawsonDfE:
...as it prevents system from working as expected and throwing error when they should be clear - this could have wider ranging issues if the existing implementation is poor. This may start flagging up lots of data issues or poor performance of the backend API

Formatting changes to amended files


ScottDawsonDfE:
The actual code changes (not formatting changes) is in the method CreateGovernorsViewModel
in the old version starting on line 788
in the new version starting on line 867

anthony9853 (approved):

PR #348, opened 2023-03-24
ids added for download pages

antk43:
ids added for all download pages

### DIFF
--- a/Web/Edubase.Web.UI/Views/Downloads/Partials/_DownloadsResults.cshtml
+++ b/Web/Edubase.Web.UI/Views/Downloads/Partials/_DownloadsResults.cshtml
@@ -47,12 +47,12 @@
 
         foreach (var group in downloadGroups)
         {
-            <h2 class="govuk-heading-m">@group.Heading</h2>
+            <h2 id="@StringUtils.ElementIdFormat(group.Heading)-heading" class="govuk-heading-m">@group.Heading</h2>
             <p>@group.Paragraph</p>
             foreach (var item in group.SubSections)
             {
                 <div class="govuk-form-group">
-                    <h3 class="govuk-heading-s">@item.Heading</h3>
+                    <h3 id="@StringUtils.ElementIdFormat(item.Heading)-heading" class="govuk-heading-s">@item.Heading</h3>
                     <div class="govuk-checkboxes">
                         @foreach (var file in item.Files)
                         {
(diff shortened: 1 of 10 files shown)


BarryGibney (approved):

Web/Edubase.Web.UI/Views/Downloads/Partials/_DownloadsResults.cshtml
```diff
@@ -47,12 +47,12 @@
...
         foreach (var group in downloadGroups)
         {
-            <h2 class="govuk-heading-m">@group.Heading</h2>
+            <h2 id="@StringUtils.ElementIdFormat(group.Heading)-heading" class="govuk-heading-m">@group.Heading</h2>
             <p>@group.Paragraph</p>
             foreach (var item in group.SubSections)
             {
                 <div class="govuk-form-group">
-                    <h3 class="govuk-heading-s">@item.Heading</h3>
+                    <h3 id="@StringUtils.ElementIdFormat(item.Heading)-heading" class="govuk-heading-s">@item.Heading</h3>
```
RogerHowellDfE, line 55: I don't know it well enough so maybe a non-issue, but I note here we're two-loops deep.

Could `item.Heading` be repeated between groups?

Could `item.Heading` be the same value as `group.Heading`?

Suggestion, use both values within the nested part:

 ```cshtml
<h3 id="@StringUtils.ElementIdFormat(group.Heading)-@StringUtils.ElementIdFormat(item.Heading)-heading" class="govuk-heading-s">@item.Heading</h3>
```

Comment (antk43) [reply Web/Edubase.Web.UI/Views/Downloads/Partials/_DownloadsResults.cshtml:55]: Hi @RogerHowellDfE thanks for spotting this.
I don't think we'd ever be in the position of this occurring as (see below) the group heading is `Establishment downloads` and the item heading is the sub group all of which have different values:
image

File: Web/Edubase.Web.UI/Views/Downloads/Partials/_DownloadsResults.cshtml [hunk 2]
```diff
@@ -64,9 +64,9 @@
                             <div class="govuk-checkboxes__item">
                                 @Html.Hidden($"{itemPrefix}{nameof(file.Item2.Tag)}", file.Item2.Tag)
                                 @Html.Hidden($"{itemPrefix}{nameof(file.Item2.FileGeneratedDate)}", file.Item2.FileGeneratedDate)
-                                <input id="@checkboxName" type="checkbox" name="@checkboxName" class="govuk-checkboxes__input" value="true"
+                                <input id="@StringUtils.ElementIdFormat(file.Item1.ToString())-checkbox" type="checkbox" name="@checkboxName" class="govuk-checkboxes__input" value="true"
                                        data-trackinglabel="@string.Concat(file.Item2.Tag,"|", file.Item2.FileGeneratedDate, "|", file.Item2.FileGeneratedDate < DateTime.Today)" />
-                                <label for="@checkboxName" class="govuk-checkboxes__label">@checkboxLabel</label>
+                                <label id="@StringUtils.ElementIdFormat(file.Item1.ToString())-checkbox-label" for="@checkboxName" class="govuk-checkboxes__label">@checkboxLabel</label>
```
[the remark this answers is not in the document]
antk43: Yes the file names are unique but of course there is a tiny risk that we end up with one being duplicated in the future. Unlikely to occur so I'm happy to leave it as-is here.

RogerHowellDfE (approved):

PR #364, opened 2023-04-25
All Minor and Patch updates

BarryGibney:
Update Minor, Patch and Major updates - first commit will be all minor and patch:
Webpack
Sass
@babel/preset-env
Core-JS
Autoprefixer
JQuery
PostCss

All major updates will be committed individually.


BarryGibney:
Note: At this point I have uploaded everything apart from Stylelint

I can't update the Stylelint packages due to errors.
I know I could update 'Stylelint-Webpack-plugin' successfully, however I have not done this update in case it has a hidden knock on effect with the other none updated Stylelint packages.
I will continue to look into this.

When completing the updates I checked after every update that I was able to successfully:
1 - open the web app
2 - sign in
3 - search for an establishment
4 - remove and add new governor 
No errors occurred.

BarryGibney:
I have checked the logs and made the updates in this order:
stylelint
stylelint-scss
stylelint-webpack-plugin
stylelint-config-standard

This has been checked and appears to be working fine - I have tested how I tested the above dependencies and they appear to be working as expected.

ScottDawsonDfE (approved):

ScottDawsonDfE:
Awaiting a check on the the webpack process running correctly

ScottDawsonDfE (commented):
Checked with @BarryGibney we need to check that webpack processes run correctly

PR #365, opened 2023-04-27
Minor version updates - working with webpack build process

BarryGibney:


BarryGibney:
Minor version updates completed together with no problems.
Major updates above completed individually
Working as expected

The below version updates produce errors in the webpack build process:

Vue (inc. vue-loader)
Stylelint (including stylelint-css) 
css-loader

ScottDawsonDfE (changes requested):
I would revert the GDS version back due to the change in how it works. While I would love to update everything, we have to consider the effort it takes to resolve the issues

PR #384, opened 2023-05-22
added new logic\enum values

anthony9853:


Web/Edubase.Services/Enums/EnumSets.cs
```diff
@@ -36,7 +36,8 @@ public static class EnumSets
             ET.FreeSchoolsAlternativeProvision,
             ET.FreeSchoolsSpecial,
             ET.StudioSchools,
-            ET.UniversityTechnicalCollege
+            ET.UniversityTechnicalCollege,
+            ET.AcademySecure16to19
```
anthony9853, line 40: This appears to not have any references in other parts of the code.  So it maybe is not used.  Talked with Mark and added this because it is a list containing academies.

Web/Edubase.Web.UI/Models/EditEstablishmentModel.cs
```diff
@@ -289,7 +289,7 @@ public GroupModel LegalParentGroup
 
         public bool IsLAMaintained => TypeId.OneOfThese(ET.CommunitySchool, ET.FoundationSchool, ET.LANurserySchool, ET.PupilReferralUnit, ET.VoluntaryAidedSchool, ET.VoluntaryControlledSchool, ET.CommunitySpecialSchool, ET.FoundationSpecialSchool);
 
-        public bool IsAcademy => TypeId.OneOfThese(ET.Academy1619Converter, ET.Academy1619SponsorLed, ET.AcademyAlternativeProvisionConverter, ET.AcademyAlternativeProvisionSponsorLed, ET.AcademyConverter, ET.AcademySpecialConverter, ET.AcademySpecialSponsorLed, ET.AcademySponsorLed, ET.FreeSchools, ET.FreeSchools1619, ET.FreeSchoolsAlternativeProvision, ET.FreeSchoolsSpecial, ET.StudioSchools, ET.UniversityTechnicalCollege, ET.CityTechnologyCollege);
+        public bool IsAcademy => TypeId.OneOfThese(ET.Academy1619Converter, ET.Academy1619SponsorLed, ET.AcademyAlternativeProvisionConverter, ET.AcademyAlternativeProvisionSponsorLed, ET.AcademyConverter, ET.AcademySpecialConverter, ET.AcademySpecialSponsorLed, ET.AcademySponsorLed, ET.FreeSchools, ET.FreeSchools1619, ET.FreeSchoolsAlternativeProvision, ET.FreeSchoolsSpecial, ET.StudioSchools, ET.UniversityTechnicalCollege, ET.CityTechnologyCollege, ET.AcademySecure16to19);
```
anthony9853, line 292: Did a bit more investigation into this.  Looks like it controls a message that is displayed in the html.  It is correct that it is an academy.  But Mark is checking if the message is correct.

Web/Edubase.Web.UI/Models/EstablishmentDetailViewModel.cs
```diff
@@ -29,7 +29,7 @@ public EstablishmentDetailViewModel(IExternalLookupService extService = null)
         {
             [(int) eLookupGroupType.SingleacademyTrust] = "Single-academy trust",
             [(int) eLookupGroupType.MultiacademyTrust] = "Academy trust",
-            [(int) eLookupGroupType.SecureSingleAcademyTrust] = "Academy trust",
+            [(int) eLookupGroupType.SecureSingleAcademyTrust] = "Secure academy trust",
```
anthony9853, line 32: This just looked like it had the wrong text in it.

Web/Edubase.Services/Enums/eLookupGroupType.cs
```diff
@@ -1,6 +1,9 @@
 
 namespace Edubase.Services.Enums
 {
+    /// <summary>
+    /// This has the same values as the database table EstablishmentGroupType so has a confusing name
+    /// </summary>
```
anthony9853, line 6: Added comment to clarify this is establishment group type and not one other the other tables.

Web/Edubase.Web.UI/Models/EstablishmentDetailViewModel.cs
CathLass: This controls the label text on the screen next to the academy name when viewing / editing the establishment

CathLass: that's one for Mark I would think - if it needs to change would it not be 'Secure single-academy trust'

anthony9853: waiting for response from Mark

Web/Edubase.Web.UI/Models/EditEstablishmentModel.cs
anthony9853: Mark to create new story with ac's for this.

anthony9853: code reverted to original value

Web/Edubase.Web.UI/Models/EstablishmentDetailViewModel.cs
anthony9853: Mark confirmed this is ok as is labelled "Academy trust"

anthony9853: original value reverted

PR #385, opened 2023-05-24
147382 Bugfix: Disabled input allows updates

Kola-Abolarinwa:
# Description
## Type of change
- [X] Bugfix

## How Has This Been Tested?
- [X] Manually Tested by Dev [@Kola-Abolarinwa ] on local environment.

Steps:
On your local machine, please follow the steps as addressed on the ticket/story:
## Checklist:
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code

### DIFF
--- a/Web/Edubase.Web.UI/Areas/Groups/Controllers/GroupController.cs
+++ b/Web/Edubase.Web.UI/Areas/Groups/Controllers/GroupController.cs
@@ -388,7 +388,7 @@ namespace Edubase.Web.UI.Areas.Groups.Controllers
                 OriginalStatusId = domainModel.StatusId,
                 UKPRN = domainModel.UKPRN.ToInteger()
             };
-            TempData["CompaniesHouseNumber"] = domainModel.CompaniesHouseNumber;
+            // TempData["CompaniesHouseNumber"] = domainModel.CompaniesHouseNumber;
             viewModel.ListOfEstablishmentsPluralName = NomenclatureService.GetEstablishmentsPluralName((GT)viewModel.GroupTypeId.Value);
 
             await PopulateEstablishmentList(viewModel.LinkedEstablishments.Establishments, id, true);
@@ -414,7 +414,7 @@ namespace Edubase.Web.UI.Areas.Groups.Controllers
             result.AddToModelState(ModelState, string.Empty);
 
             await PopulateSelectLists(viewModel);
-            if (TempData.TryGetValue("CompaniesHouseNumber", out var value)&& viewModel.CompaniesHouseNumber != value.ToString())
+            if (TempData.TryGetValue("companiesHouseNumber", out var value))
                 viewModel.CompaniesHouseNumber = value.ToString();
 
             if (viewModel.CanUserEditStatus)
--- a/Web/Edubase.Web.UI/Areas/Groups/Controllers/GroupController.cs
+++ b/Web/Edubase.Web.UI/Areas/Groups/Controllers/GroupController.cs
@@ -388,6 +388,7 @@ namespace Edubase.Web.UI.Areas.Groups.Controllers
                 OriginalStatusId = domainModel.StatusId,
                 UKPRN = domainModel.UKPRN.ToInteger()
             };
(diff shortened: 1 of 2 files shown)


CathLass (commented):
Needs a discussion and decision as to wider issues - see ticket.

Web/Edubase.Web.UI/Areas/Groups/Views/Group/EditDetails.cshtml
```diff
@@ -140,13 +140,14 @@
...
+                        TempData["companiesHouseNumber"] = @Model.CompaniesHouseNumber;
                         <div class="govuk-form-group">
                             <label class="govuk-label" for="companiesHouseNumber">
                                 Company number<br />
                                 (also known as Companies House (CH) number)
                             </label>
                             <input type="text" value="@Model.CompaniesHouseNumber" class="govuk-input" disabled="disabled" id="companiesHouseNumber" />
-                            @Html.HiddenFor(x => x.CompaniesHouseNumber)
+                            @* @Html.HiddenFor(x => x.CompaniesHouseNumber) *@
                         </div>
```
RogerHowellDfE, line 151: I do not fully understand the code changes here and how it helps with the access control - maybe we could have a call and talk through it? :) 

I see that the other fields have an access / permission check around it like these examples below -- have we tried an approach like this to control if the input is visually displayed as editable vs disabled/hidden?

It may also be that the backend API and/or permissions within the database need to be updated also.

```cshtml
                    @if (Model.CanUserEditUkprn)
                    {
                        // editable field
                    }
                    else 
                    {
                        // disabled or hidden/"not visible" input
                    }

                    @if (Model.CanUserEditStatus)
                    {
                        // editable field
                    }
                    else 
                    {
                        // disabled or hidden/"not visible" input
                    }

                    @if (Model.CanUserCloseAndMarkAsCreatedInError)
                    {
                        // editable field
                    }
                    else 
                    {
                        // disabled or hidden/"not visible" input
                    }
```

Kola-Abolarinwa: Sure Roger, we can have a call about it. It already uses a similar code on the front end 
so the code change is actual in one of those blocks.

Web/Edubase.Web.UI/Areas/Groups/Controllers/GroupController.cs
```diff
@@ -388,6 +388,7 @@ public async Task<ActionResult> EditDetails(int id)
                 OriginalStatusId = domainModel.StatusId,
                 UKPRN = domainModel.UKPRN.ToInteger()
             };
+            // TempData["CompaniesHouseNumber"] = domainModel.CompaniesHouseNumber;
```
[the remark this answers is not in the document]
Kola-Abolarinwa: Sorry, my bad

Web/Edubase.Web.UI/Areas/Groups/Views/Group/EditDetails.cshtml
RogerHowellDfE: @Kola-Abolarinwa has kindly spared some time to walk through this, thank you! 

Notes to self / anybody else following along:
- `TempData` is provided by MVC, where values are scoped to the current user's session and the value is removed on the first read of this data
- Potential bug possible if editing multiple groups within multiple tabs (`TempData` is global for the session, not per browser tab) - planned fix for this specific field is to suffix the key with the group`s UID to prevent 

**Once the key has been updated this implementation appears to work as desired to prevent user-supplied values entering into the system for the companies house number field, while also making sure that it is available to submit to the backend API.**

Additional database changes are required for belt+braces, and to address the underlying issue of the wrong permissions being applied to some users within the database (not visible due to the C# code disregarding any database/API permissions and having hard-coded permissions logic within the view layer).

A (much!) broader question is how/whether to address this more broadly than companies house number, and rearchitecting things to have a single source of truth for all authorisation checks rather than having it spread across multiple layers.

FYI @CathLass @ScottDawsonDfE @anthony9853

ScottDawsonDfE:
I have approved, details of why on the story on ADO

PR #388, opened 2023-05-31
YCS added to the Tools for managing requests

FilAntoniadis:


RogerHowellDfE (approved):
Looks good to me :) 

- This makes the tool link visible on the tools page for YCS users
- The list of change requests appears to be filtered to include only changes made to 16-19 (logging in as a backoffice user shows an additional 22 change requests for other establishment types)
- Approving a change appears to work as a backoffice user
- Approving a change appears to work as a YCS user
- Attempting to approve a change as a YCS user that they would not normally have permission to approve/reject is (correctly) denied
- Rejecting as a change appears to work as a YCS user

PR #390, opened 2023-06-06
change config for lookups API

CathLass:
**I've made modifications to this since getting approvals so I tried to reset the approvals but failed!**

Move back to getting lookups from the java API rather than the c# replacement API. The c# replacement API has not been released to production and we are adding functionality to the java lookups API for 16-19, hence going back to using that API for lookups.

This change is related to previous PRs https://github.com/DFE-Digital/get-information-about-schools/pull/237 and https://github.com/DFE-Digital/get-information-about-schools/pull/340 where the lookups API config was split out from the texuna API config to allow different APIs for each and the lookups API was pointed to the new c# replacement API

### DIFF
--- a/Web/Edubase.Web.UI/Web.config
+++ b/Web/Edubase.Web.UI/Web.config
@@ -165,8 +165,8 @@
     <add key="api:Username" value="rest-api-user" />
     <add key="api:Password" value="" />
 
-    <add key="LookupApiBaseAddress" value="https://s158d01-func-dd-restapi.azurewebsites.net/api/" />
-    <add key="LookupApiUsername" value="" />
+    <add key="LookupApiBaseAddress" value="https://app-api-t1dv-edubase.azurewebsites.net/edubase/rest/" />
+    <add key="LookupApiUsername" value="rest-api-user" />
     <add key="LookupApiPassword" value="" />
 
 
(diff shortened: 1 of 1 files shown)


RogerHowellDfE (approved):

RogerHowellDfE:
+1 from me for this specific PR

RogerHowellDfE:
As a broader comment, shall we remove all of these default URLs? 

My personal view is that this is (and several other config options are) environment-specific configuration and there is no reasonable default that belongs in git. If we have a default for local development, then let's include that default within the local configuration supplied during onboarding 👍

- Each environment (except dev?) has this overridden anyway for the relevant API URL it should be connecting to (and should be supplied as part of the infra setup/teardown, not via the application)

- For local development, we override the username/password etc anyway, so it is trivial to also supply the API URL (and should be supplied as part of the onboarding steps/configs like we do for credentials)

If we agree this is the way to go (possibly for a discussion outside of this PR?), personally I would be inclined to accept this PR to ensure this specific API gets sorted, and then do a separate PR with bulk removal of default URLs en-masse.

anthony9853 (approved):

ScottDawsonDfE (approved):

PR #394, opened 2023-06-09
field is now only visible for non 16-19 groups

anthony9853:
validation is only uses for non 16-19 groups

### DIFF
--- a/Web/Edubase.Web.UI/Areas/Groups/Controllers/GroupController.cs
+++ b/Web/Edubase.Web.UI/Areas/Groups/Controllers/GroupController.cs
@@ -548,6 +548,12 @@ namespace Edubase.Web.UI.Areas.Groups.Controllers
 
             if (ModelState.IsValid)
             {
+                //required validation of field GroupId is now only done for groups that are not 16-19
+                if (academyTrustRoute != "secure-academy-trust" && string.IsNullOrWhiteSpace(viewModel.GroupId))
+                {
+                    ModelState.AddModelError("GroupId", "Please enter a group ID");
+                }
+
                 var validationEnvelope = await _groupWriteService.ValidateAsync(dto, User);
                 if (validationEnvelope.HasErrors)
                 {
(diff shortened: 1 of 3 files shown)


Web/Edubase.Web.UI/Areas/Groups/Controllers/GroupController.cs
```diff
@@ -548,6 +548,12 @@ public async Task<ActionResult> SaveNewAcademyTrust(CreateAcademyTrustViewModel
 
             if (ModelState.IsValid)
             {
+                //required validation of field GroupId is now only done for groups that are not 16-19
```
anthony9853, line 551: Had another look at this and think this is probably an ok way to do this.  But maybe this is the wrong place in the method to add the code.

CathLass: FluentValidation would be better but it looks like you've followed a pattern already used in the code and it's probably not the time to be changing all the validation on the page. I think it is in the wrong place though - I'd put it as early in the method as you can

Web/Edubase.Web.UI/Areas/Groups/Models/CreateEdit/CreateAcademyTrustViewModel.cs
```diff
@@ -47,7 +47,8 @@ public CreateAcademyTrustViewModel(CompanyProfile companyProfile, IEnumerable<Se
         [Required(ErrorMessage = "Please select the trust type"), DisplayName("Trust type")]
         public int? TypeId { get; set; }
 
-        [Required(ErrorMessage = "Please enter a group ID"), DisplayName("Group ID")]
+        //[Required(ErrorMessage = "Please enter a group ID"), DisplayName("Group ID")]
+        //added validation to the controller as this now depends on if the group is 16-19 secure
```
CathLass, line 51: remove comment

CathLass (changes requested):

Web/Edubase.Web.UI/Areas/Groups/Controllers/GroupController.cs
anthony9853: Moved the if statement up the method.  As you say we don't have time to restructure it now...

Web/Edubase.Web.UI/Areas/Groups/Models/CreateEdit/CreateAcademyTrustViewModel.cs
anthony9853: sorry are you saying to remove this comment?

CathLass (approved):

PR #410, opened 2023-07-03
16 19 secure/145544 governance view

CathLass:
Governance view - fix the naming of the governor roles on the views
Removes the conversion of an enum to a string - the current conversion code couldn't cope with the new governor role names having '-' e.g 'Member - indiviudal'

The nomenclature service method and associated tests aren't needed now - they're just there to show that I had the existing behaviour covered by tests before I replaced the code. Not sure what to do about that - maybe another PR to remove the unwanted code afterwards.

https://dfe-ssp.visualstudio.com/s158-Get-Information-About-Schools/_workitems/edit/145544

### DIFF
--- a/Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
+++ b/Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
@@ -112,8 +112,6 @@ namespace Edubase.Web.UI.Areas.Governors.Models
             foreach (var role in roles)
             {
                 var equivalantRoles = RoleEquivalence.GetEquivalentToLocalRole(role).Cast<int>().ToList();
-                //var pluralise = !EnumSets.eSingularGovernorRoles.Contains(role);
-
 
                 var grid = new GovernorGridViewModel($"{GovernorRoleNameFactory.Create(role, eTextCase.SentenceCase, true)}{(isHistoric ? " (in past 12 months)" : string.Empty)}")
                 {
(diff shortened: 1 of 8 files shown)


Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
```diff
@@ -111,18 +112,18 @@ private void CreateGrids(GovernorsDetailsDto dto, IEnumerable<GovernorModel> gov
             foreach (var role in roles)
             {
                 var equivalantRoles = RoleEquivalence.GetEquivalentToLocalRole(role).Cast<int>().ToList();
-                var pluralise = !EnumSets.eSingularGovernorRoles.Contains(role);
+                //var pluralise = !EnumSets.eSingularGovernorRoles.Contains(role);
```
RogerHowellDfE, line 115: This commented line conditionally flags if the role name should be plural or not, while the replacement sets this to permanently true (line 118).

If we're moving to having all roles plural on this page, let's remove this line?
Else let's uncomment this line and re-introduce the variable on line 118?

RogerHowellDfE:
Minor comment re: conditionally making the name plural (or not), but otherwise seems to work well when running this locally

Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
CathLass: Removed commented-out code. The logic as to whether a particular governor role should be pluralised has been moved to the GovernorRoleNameFactory

BarryGibney (approved):

RogerHowellDfE (approved):

CathLass:
Re: outstanding comment - I verified with Roger that the behaviour is the same as before 
The logic was two-stage before:
1.determine whether that role should be 'pluralised'
2. call the method with pluralise set accordingly

Now the factory deals with all that logic

Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
RogerHowellDfE: Addendum - I had a misunderstanding here which @CathLass has kindly corrected me on: the boolean value is `pluraliseLabelIfApplicable`, and the conditional logic is within that method.

PR #421, opened 2023-07-17
move API call inside if statement - call it only if needed

CathLass:
Small refactor, as I spotted it


RogerHowellDfE (approved):
Nice spot :)

PR #459, opened 2023-10-02
Add ids for cancel button and proceed

JaswiniP:


JaswiniP:
174529-- Cancel button

174545-- proceed button

RogerHowellDfE (approved):
I initially worried that this may lead to duplicated IDs on pages with the warning panel included/embedded within a page which also has a `#proceed-button` element, but Jaswini has checked this locally and all appears good there so marking as approved.

PR #482, opened 2023-11-23
178552 add anti csrf tokens

anthony9853:


RogerHowellDfE:
Here are the areas I have specifically looked at while reviewing this draft PR -- overall looks good to me, just a few uses of raw `<form>` tags to also add tokens too 🙂 

- [X] ✅ **Places where `BeginRouteForm` / `Html.BeginRouteForm` helpers are used**
  - All `FormMethod.Post` has the crsf token
  - The `FormMethod.Get` ones _(seemingly reasonably)_ do not have csrf tokens
- [ ] ❓ **Places where raw HTML `<form` / `</form` is used**
  - _(specifically these regexes, allowing for whitespace: `<\s*form\b` or `<\s*/\s*form\b`)_
  - A mix of get and post action raw forms
  - There are some additional opportunities here to insert csrf tokens 
- [X] ✅ **Places where `BeginForm` / `Html.BeginForm` helpers are used**
  - All `FormMethod.Post` has the crsf token
  - The `FormMethod.Get` ones _(seemingly reasonably)_ do not have csrf tokens
- [X] ✅ **Places where `FormMethod.Post` is used**
  - Overlaps with above helper methods
  - All appear to have CSRF tokens

FYI @anthony9853

RogerHowellDfE:
Just spent some time looking at the controllers

- [] ✅ **Places where `HttpPost` attribute is used - "normal" requests**
  - Appears to be fine, with each `HttpPost` having a corresponding `ValidateAntiForgeryToken`
- [] ❓ **Places where `HttpPost` attribute is used - "ajax/js" requests**
  - Typically no corresponding `ValidateAntiForgeryToken` attributes
  - Some do not (e.g., `HomeController.cs` (cookie choices ajax), `GroupApiController.cs` and other `...ApiController` classes) which may benefit from being double checked/reviewed (+- input from other devs as I do not currently have a strong view)
  - ... It is not immediately obvious to me if these should or should not (I understand these are typically called from the JavaScript, which introduces some extra complexity / additional places to check re: insertion of CSRF tokens perhaps?)
  - _(if explicitly do not need / should not have validation in these places, let's add a comment to this effect perhaps?)_

anthony9853:
@RogerHowellDfE ready for review.  Had to update it as dev has moved on since put probably ok.

anthony9853:
@RogerHowellDfE spotted some changes after the update from dev.  looks like its just whitespace.

RogerHowellDfE (approved):
- Uses of `<form>`
  - Default `<form>` method is `GET`, unless specifically overridden
  - All uses of `<form>` either use default or explicit `GET` method - no uses of `POST` seen by me
- API routes called from ajax/js
  - None of the `HttpPost` routes validate CSRF tokens, thus shouldn't break / change from this PR (running it locally seemed to work fine for the handful I checked)
  - I understand a new work item has been created to investigate/address these in follow-up work (caveat: couldn't find it)

Given this, I don't see a reason not to merge this PR 👍

ScottDawsonDfE (approved):

PR #484, opened 2023-11-29
Revert "Merge pull request #480 from DFE-Digital/rh/178958--rename-fscp"

RogerHowellDfE:
Following some debate today and yesterday it appears additional discussion is required. 

This, therefore, reverts PR #480 to rollback this change until those discussions can take place.


RogerHowellDfE:
Resolution achieved via different approach. Closing this revert PR.

PR #490, opened 2024-01-09
BG/184252-Remove_excessive_retries

BarryGibney:
Initial commit - OSPlacesApiService changed in the web.config to disable retries

### DIFF
--- a/Web/Edubase.Services/IntegrationEndPoints/AzureMaps/AzureMapsService.cs
+++ b/Web/Edubase.Services/IntegrationEndPoints/AzureMaps/AzureMapsService.cs
@@ -5,6 +5,7 @@ namespace Edubase.Services.IntegrationEndPoints.AzureMaps
     using System.IO;
     using System.Linq;
     using System.Net.Http;
+    using System.Threading;
     using System.Threading.Tasks;
     using Common;
     using Edubase.Common.Spatial;
@@ -48,15 +49,10 @@ namespace Edubase.Services.IntegrationEndPoints.AzureMaps
                 .Select(x => TimeSpan.FromSeconds(int.Parse(x)))
                 .ToArray();
 
-            if (retryIntervals.Length == 0)
-            {
-                return new[] { TimeSpan.FromSeconds(2) };
-            }
-
             return retryIntervals;
         }
 
-        public async Task<PlaceDto[]> SearchAsync(string text, bool isTypeahead)
+        public async Task<PlaceDto[]> SearchAsync(string text, bool isTypeahead, CancellationToken cancellationToken = default)
         {
             text = text.Clean();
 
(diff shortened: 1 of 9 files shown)


BarryGibney:
The ticket suggested I use a disable feature for the retries, it didn't fully explain how this should be implemented, unfortunately I didn't get any questions answered - so this is how I have done it:

You can still use the retries, say 1,2,2,4
If you want one timeout for e.g 3 seconds without retrying you can do that ...so 3
If you want to disable the retries, this will then go to whatever default value is set

Web/Edubase.Services/IntegrationEndPoints/OSPlaces/OSPlacesApiService.cs
```diff
@@ -25,8 +25,29 @@ public class OSPlacesApiService : IOSPlacesApiService
 
         private static Policy CreateRetryPolicy()
         {
-            var retryIntervalSettings = ConfigurationManager.AppSettings["OSPlacesApiServices_RetryIntervals"].Split(',');
-            var retryIntervals = retryIntervalSettings.Select(int.Parse).ToArray();
+            var defaultRetriesValue = new[] { 2 };
+            var retryEnabled = ConfigurationManager.AppSettings["OSPlacesApiServices_TimeoutIntervals_Enabled"] == "1";
+            var retryIntervalSettings = ConfigurationManager.AppSettings["OSPlacesApiServices_TimeoutIntervals"];
```
RogerHowellDfE, line 30: This behaviour is specifically about the intervals between retry attempts? 

If so, I'd suggest not including `TimeoutIntervals` in the name as this implies it is specifically about the timeout (how long the request is allowed to run for, as opposed to the gaps between the requests).

Similarly, whether the retries are enabled or not?

RogerHowellDfE:
I'll follow up in chat with some other questions / comments / suggestions.

Web/Edubase.Services/IntegrationEndPoints/OSPlaces/OSPlacesApiService.cs
anthony9853: Probably should be RetryIntervals.  Does this allow an empty set of intervals as no retries at all?

Web/Edubase.Services/IntegrationEndPoints/AzureMaps/AzureMapsService.cs
```diff
@@ -26,12 +26,34 @@ public class AzureMapsService : IAzureMapsService
...
+            var retryIntervals = csvSeconds.Split(',')
+                .Select(x => x.Trim())
+                .Where(x => int.TryParse(x, out var seconds) && seconds >= 0)
+                .Select(x => TimeSpan.FromSeconds(int.Parse(x)))
+                .ToArray();
+
+            if (retryIntervals.Length == 0)
+            {
+                return new[] { TimeSpan.FromSeconds(2) };
+            }
```
RogerHowellDfE, line 54: For what reason are we defaulting to two attempts, with 2-seconds between the attempts?

Apologies if I've missed something with this!

BarryGibney: we shouldn't be - this should be applied to default timeouts not retries

RogerHowellDfE (approved):

PR #491, opened 2024-01-15
BG/191418 - Create disable feature for retries as well as creating a value we can change for timeouts

BarryGibney:
Create disable feature for retries as well as creating a value we can change for timeouts

### DIFF
--- a/Web/Edubase.Web.UI/Web.config
+++ b/Web/Edubase.Web.UI/Web.config
@@ -285,8 +285,10 @@
     <add key="SfbClient_Timeout" value="2" />
     <add key="OSPlacesApiServices_RetryIntervals" value="1,2,2,4"/>
     <add key="AzureMapService_RetryIntervals" value="1,2,2,4"/>
-    
 
+    <add key="RetryTimeoutSettings" value="10"/>
+
+    
     <!--
       ACCESS TO INTERACT WITH DATA HELD BY GIAS
 
--- a/Web/Edubase.Web.UI/Web.config
+++ b/Web/Edubase.Web.UI/Web.config
@@ -283,11 +283,10 @@
     <add key="LookupClient_Timeout" value="10" />
     <add key="FscpdClient_Timeout" value="2" />
     <add key="SfbClient_Timeout" value="2" />
+    <add key="AzureMapService_Timeout" value="2"/>
+    <add key="OSPlacesApiServices_Timeout" value="10"/>
     <add key="OSPlacesApiServices_RetryIntervals" value="1,2,2,4"/>
     <add key="AzureMapService_RetryIntervals" value="1,2,2,4"/>
-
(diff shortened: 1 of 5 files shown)


anthony9853 (approved):

Web/Edubase.Web.UI/Web.config
```diff
@@ -285,8 +285,11 @@
     <add key="SfbClient_Timeout" value="2" />
     <add key="OSPlacesApiServices_RetryIntervals" value="1,2,2,4"/>
     <add key="AzureMapService_RetryIntervals" value="1,2,2,4"/>
-    
 
+    <add key="AzureMapServiceTimeoutKey" value="2"/>
+    <add key="OSPlacesApiServicesTimeoutKey" value="10"/>
```
RogerHowellDfE, line 290: What is the rationale for these being named differently to and being added separately to the `xxx_Timeout` timeout values above please?

RogerHowellDfE:
@BarryGibney - Just looking at the other timeouts, they're done by setting `Timeout` when calling `new HttpClient` (next to specifying the `BaseAddress`).

Perhaps we could follow that convention in this PR rather than using the Polly timeout?

I suppose some will depend on / spill over into @anthony9853 's work to make our use of `HttpClient` consistent, so maybe he has a view (e.g., perhaps he intends to move the others to use Polly too which would also result in it being consistent)?


https://github.com/DFE-Digital/get-information-about-schools/blob/b543a5e604ae40a88149b9311871fa061c6ac619/Web/Edubase.Web.UI/App_Start/IocConfig.cs#L222-L229

https://github.com/DFE-Digital/get-information-about-schools/blob/b543a5e604ae40a88149b9311871fa061c6ac619/Web/Edubase.Web.UI/App_Start/IocConfig.cs#L243-L250

https://github.com/DFE-Digital/get-information-about-schools/blob/b543a5e604ae40a88149b9311871fa061c6ac619/Web/Edubase.Web.UI/App_Start/IocConfig.cs#L264-L271

https://github.com/DFE-Digital/get-information-about-schools/blob/b543a5e604ae40a88149b9311871fa061c6ac619/Web/Edubase.Web.UI/App_Start/IocConfig.cs#L288-L301

Web/Edubase.Web.UI/Web.config
BarryGibney: To identify these values differently from the other timeouts, since these timeouts also had retries - I will move them back with the others and rename them, as you can also compare them to the retry list (if there is a lot more added it might be a good idea to identify them this way?)

BarryGibney:
@RogerHowellDfE 
Might be a bit tricky from what I understand.
I am setting a timeout for each retry attempt - which as far as I am aware, the httpclient is more for simple requests (so would set the timeout value to the lot instead of each individual retry)?

PR #501, opened 2024-02-05
BG/193913_add Shared governance professional to Group/Establishment level

BarryGibney:
Correct an error that appears when adding a 'Shared governance professional - group' and a 'Governance professional to a MAT'


BarryGibney:
The original code is ensuring only a single governance can be added.
This ticket fixes the bug as a temporary way until we receive a full list of what governance can be active at the same time as others, then a ticket will need creating to implement all scenarios

anthony9853 (approved):

PR #502, opened 2024-02-06
Bg/193462 Character count is missing

BarryGibney:
Character count missing in notification banner

### DIFF
--- a/Web/Edubase.Web.UI/Views/Notifications/EditBanner.cshtml
+++ b/Web/Edubase.Web.UI/Views/Notifications/EditBanner.cshtml
@@ -204,14 +204,6 @@
                     </div>
                 </div>
 
-                <div class="govuk-character-count" data-module="govuk-character-count" data-maxlength="500">
-                    <div class="govuk-form-group">
-                        <div id="with-hint-hint" class="govuk-hint">
-                            Keep the message clear, concise and user focused. The maximum number of characters permitted is 500. This includes spaces.
-                        </div>
-                    </div>
-                </div>
-
                 <details class="govuk-details govuk-!-margin-bottom-4" data-module="govuk-details">
                     <summary class="govuk-details__summary">
                         <span class="govuk-details__summary-text">
--- a/Web/Edubase.Web.UI/Views/Notifications/EditBanner.cshtml
+++ b/Web/Edubase.Web.UI/Views/Notifications/EditBanner.cshtml
@@ -231,101 +231,88 @@
                     </div>
                 </details>
                 <div class="govuk-character-count" data-module="govuk-character-count" data-maxlength="500">
-                        <div class="govuk-form-group @Html.ValidationCssClassFor(x => x.Content)">
-                            @Html.ValidationMessageFor(x => x.Content, null, new { @class = "govuk-error-message" })
-                            @Html.TextAreaFor(x => x.Content, new { @class = "govuk-textarea govuk-js-character-count govuk-textarea entry-content", @id = "with-hint", @name = "withHint", @rows = "5" })
-                            <div id="with-hint-info" class="govuk-hint govuk-character-count__message">
-                                You can enter up to 500 characters
-                            </div>
+                    <div class="govuk-form-group @Html.ValidationCssClassFor(x => x.Content)">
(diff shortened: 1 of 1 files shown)


Web/Edubase.Web.UI/Views/Notifications/EditBanner.cshtml
```diff
@@ -223,9 +222,14 @@
...
-                    @Html.TextAreaFor(x => x.Content, new { @class = "govuk-textarea entry-content", rows = "6", cols = "40", maxlength = "500" })
+                <div class="govuk-character-count" data-module="govuk-character-count" data-maxlength="500">
+                    <div class="govuk-form-group @Html.ValidationCssClassFor(x => x.Content)">
+                        @Html.ValidationMessageFor(x => x.Content, null, new { @class = "govuk-error-message" })
+                        @Html.TextAreaFor(x => x.Content, new { @class = "govuk-textarea govuk-js-character-count govuk-textarea entry-content", @id = "with-hint", @name = "withHint", @rows = "5", maxlength = "500" })
+                        <div id="with-hint-info" class="govuk-hint govuk-character-count__message">
+                            You can enter up to 500 characters
+                        </div>
+                    </div>
                 </div>
```
RogerHowellDfE, line 233: Nice work using the [GDS character count component](https://design-system.service.gov.uk/components/character-count/) for this! 👍 

**Comments (needs):**
- The text area was missing an `aria-describedby`, which meant that e.g., screen-readers would not be able to speak aloud the characters remaining
  - In the suggestion below, I have added the label but made it `govuk-visually-hidden` so that it is available to screenreaders but won't affect/distract visual users (debatable if it should be hidden, but I figure it's least intrusive option most likely to be accepted)
  - If we were to have a "hint" we should also include that in the `aria-describedby` too
- The `@name` isn't needed (and seems to get overridden anyway, therefore keeping it would be misleading)
  - .. therefore I've removed it from the suggested edit below

**Comments (nice-to-haves):**
- Duplicated `govuk-textarea` class
  - This should be removed
- I'm not sure what `entry-content` does and if we can remove it
  - It was already there so presumably won't hurt to leave it :) 
- Original had 6 rows, current diff has it less tall with only 5 rows
- I've changed the `id` below
  - Originally I did it so that the ID was more relevant to the page/content, but then I remembered that if we don't keep the IDs synchronised with the auto-generated ID then it breaks the error summary clickable links (and the issues we had before around these disappearing if it can't find the linked input ID?)
  - There's is a way to do this dynamically like `@Html.GenerateId(x => x.Content)` or something like that, but I couldn't find it just now (that would then mean being able to remove the `@id = ""` bit too)
- There are multiple references to `500` as the character limit
  - I suggest adding this as a `const` to `Edubase.Web.UI.Models.Notifications.NotificationsBannerViewModel` (the class which does the server-side validation - see the `MaxLength` attribute) and referencing that, so we can keep all of the character count references consistent


```cshtml
                <div class="govuk-character-count" data-module="govuk-character-count" data-maxlength="500">
                    <div class="govuk-form-group @Html.ValidationCssClassFor(x => x.Content)">
                        <label class="govuk-label govuk-label--s govuk-visually-hidden" for="Content">
                            What is the notification banner message?
                        </label>
                        @Html.ValidationMessageFor(x => x.Content, null, new {@class = "govuk-error-message"})
                        @Html.TextAreaFor(
                            x => x.Content,
                            new
                            {
                                @class = "govuk-textarea govuk-js-character-count entry-content",
                                @id = "Content",
                                @rows = "5",
                                maxlength = "500",
                                aria_describedby = "Content-info",
                            }
                        )
                        <div id="Content-info" class="govuk-hint govuk-character-count__message">You can enter up to 500 characters</div>
                    </div>
                </div>
```

BarryGibney: Thanks for reviewing this Roger - I have made the changes suggested. 
I have also changed the 5 rows from the design service HTML to the original 6.
I haven't included the const to NotificationsBannerViewModel as I believe this may be a bit more involved (I am not sure it would be easily accessible to the javascript)

RogerHowellDfE (approved):

PR #518, opened 2024-02-26
Bg/bug 135244 add shared gov pro

BarryGibney:
Currently due to how the code works, the Shared Gov Pro - Establishment will return as a Gov Pro To An Individual Academy Or Free School.

### DIFF
--- a/Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
+++ b/Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
@@ -114,22 +114,6 @@ namespace Edubase.Web.UI.Areas.Governors.Models
 
             foreach (var role in roles)
             {
-                //var roleName = role;
-                //if (roleName == GR.GovernanceProfessionalToAnIndividualAcademyOrFreeSchool)
-                //{
-                //    foreach (var item in dto.CurrentGovernors)
-                //    {
-                //        if (item.RoleId == 20)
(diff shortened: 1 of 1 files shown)


Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
```diff
@@ -130,6 +130,21 @@ public GovernorsGridViewModel()
...
+                // this code changes the on screen title to display the correct title for the roleID
+
+                if (role == GR.GovernanceProfessionalToAnIndividualAcademyOrFreeSchool)
+                {
+                    foreach (var item in dto.CurrentGovernors)
+                    {
+                        if (item.RoleId == 20)
+                        {
+                            governorRoleNameTitle = "Shared governance professional establishment";
+                        }
```
RogerHowellDfE, line 144: For this, could we use the enums (instead of the int) and naming service (instead of the raw string) to provide these values rather than hard-coding them in?

BarryGibney: Can change to: if (item.RoleId == (int)GR.Establishment_SharedGovernanceProfessional) if that makes it clearer? - If you mean name change in GovernorRoleNameFactory there isn't a 'Shared governance professional establishment' (there is one with a dash in but that isn't what is expected)

anthony9853 (approved):

PR #521, opened 2024-03-07
Historic title display for Shared Governance Professional (establishment)

BarryGibney:
Historic title display for Shared Governance Professional (establishment)

### DIFF
--- a/Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
+++ b/Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
@@ -109,24 +109,18 @@ namespace Edubase.Web.UI.Areas.Governors.Models
         {
             var roles = dto.ApplicableRoles.Where(role =>
             {
+                if (!EnumSets.eSharedGovernorRoles.Contains(role))
                 {
-                    if (!EnumSets.eSharedGovernorRoles.Contains(role))
-                    {
-                        return true;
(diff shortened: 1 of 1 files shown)


RogerHowellDfE (approved):

Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
```diff
@@ -106,11 +107,24 @@ public GovernorsGridViewModel()
...
+                    return true;
+                }
+
+                var localEquivalent = RoleEquivalence.GetLocalEquivalentToSharedRole(role);
+                if (localEquivalent != null)
+                {
+                        if (!dto.ApplicableRoles.Contains(localEquivalent.Value))
+                        {
+                            return true;
+                        }
```
RogerHowellDfE, line 123: Missing the `else { return false; }` in this refactor?

RogerHowellDfE (changes requested):

Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
BarryGibney: the else { return false; } restricts the roles so the additional governor is not seen - the way it is now is the opposite and allows more than we need, so this has been refactored

RogerHowellDfE (approved):

PR #525, opened 2024-03-19
BG/Bug_201204_Overwriting

BarryGibney:
This is where local chair and shared chair were overwriting each other


RogerHowellDfE (approved):
This resolves the bug/issue of the "shared" and "local" equivalents (wrongly) not being detected as being equivalent thus rejected -- we discussed this on a call and saw this fix working ✅

PR #533, opened 2024-04-18
Test Locators on establishment details page

lpotter-dfe:
Added in a couple of locators on the URN number at the top of the page and a locator for the establishment parent group link.


RogerHowellDfE (approved):
ID's appear to be unique and HTML appears to be correct -- ✅

anthony9853 (approved):

PR #537, opened 2024-04-23
Show a textual explanation/description that there are no governance roles available, instead of an empty drop-down select list.

RogerHowellDfE:
Currently, in some circumstances where there are no governance roles available/permitted to be added to an establishment/group, we display an empty drop-down list and this has led to some confusion.

This PR, instead, shows a textual explanation/description that there are no governance roles available.

#203275

image


anthony9853 (approved):

RogerHowellDfE:
Rebased due to merge conflicts.

PR #542, opened 2024-05-15
Do not clear gov role on error

RogerHowellDfE:
Where we deny the user creating a new governance role record because a person in that role already exists (e.g., chair of trustees, accounting officer, and the new governance professional roles), the drop-down box currently defaults to whichever entry is top of the list.

During governance professional testing, it is felt this is misleading as it initially looks like whatever value is first in the list already has an appointee.

Additionally, WCAG 2.2 requires form fields to not be cleared when validating user input.
https://design-system.service.gov.uk/patterns/validation/#wcag-edit-to-correct-errors

image

For this reason, this PR has a basic implementation which causes the drop-down to pre-select whichever role the user submitted.


RogerHowellDfE:
I have rebased this branch onto the `dev` branch, and have re-targeted this PR to the `dev` branch (was previously based on/targeted at the governance professional feature/epic branch).

Once the pipeline checks are through, it'll be ready for review!

BarryGibney (approved):

PR #551, opened 2024-06-19
Minimal PR making RSC region respect display policy

RogerHowellDfE:
This is a minimal PR is targeting dev.

I have specifically opted to not to add similar checks for the remaining location fields to avoid any unwanted issues (e.g., if permissions are not correct on the other fields, fields may inadvertently disappear -- this will be the topic of separate follow-up work).

Note this is a follow-up to #550, targeting the main/master branch.


RogerHowellDfE:
Closing, as alternative approach taken in a separate PR.

PR #578, opened 2024-08-20
222574 remove applicable role filter

anthony9853:
displayed governor roles that are now determined from the roles in the api response and not just the applicable roles

### DIFF
--- a/Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
+++ b/Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
@@ -109,9 +109,7 @@ namespace Edubase.Web.UI.Areas.Governors.Models
         private void CreateGrids(GovernorsDetailsDto dto, IEnumerable<GovernorModel> governors, bool isHistoric,
             int? groupUid, int? establishmentUrn)
         {
-            var currentRoles = dto.CurrentGovernors.Select(g => (GR) g.RoleId).Distinct();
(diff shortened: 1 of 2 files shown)


Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
```diff
@@ -109,7 +109,9 @@ public GovernorsGridViewModel()
         private void CreateGrids(GovernorsDetailsDto dto, IEnumerable<GovernorModel> governors, bool isHistoric,
             int? groupUid, int? establishmentUrn)
         {
-            var roles = dto.ApplicableRoles.Where(role =>
+            var allRoles = governors.Select(g => (GR) g.RoleId).Distinct();
+
+            var roles = allRoles.Where(role =>
```
RogerHowellDfE, line 114: If we remove the `applicableRoles` completely, the sections which say something to the effect of "there are no appointees for this role" will not show (instead, the headings will display only if there is a record of that type recorded).

This isn't necessarily a bad thing, but it is a change in behaviour.

To retain current behaviour, we could union with the applicable roles perhaps?

RogerHowellDfE (commented):
There is currently a failing test. 

Based on a quick glance over, it is currently unclear to me if this is just a test setup issue.

image

Web/Edubase.Web.UI/Areas/Governors/Models/GovernorsGridViewModel.cs
anthony9853: think all this is fixed now

BarryGibney (approved):

PR #608, opened 2024-11-12
Copy updated logic to historic section too

RogerHowellDfE:
The logic to include additional types of "shared governor" was previously updated for the "current" governor buttons, but it looks like the "historical" governor section was missed.

This PR resolves this.


RogerHowellDfE:
The current PR is invalid -- this code section is only for the _group-level_ governors, whereas the current draft applies to _all_ shared/sharing governors (incorrect).

Potentially some additional work to do, to verify if adding a third role is required.

PR #624, opened 2024-11-26
BG/239685 - HttpException accessing parent directory

BarryGibney:


anthony9853 (commented):
Had a look at this and it looks ok.  Will have to have a chat to discuss how this is working in more detail.

anthony9853 (approved):

PR #847, opened 2026-03-10
Fix to duplicate historic govs including test

antk43:
**Overview**
Recent changes in the Java to display historic shared local governors and local governors correctly within an establishment that is part of a trust has resulted in duplicate governors displaying. 
The duplicate governors had the same GID but had different roles due to the governance sharing option being changed.
Investigations came across a role equivalence function which we're not totally certain of its purpose as discussed here: https://dfe-ssp.visualstudio.com/s158-Get-Information-About-Schools/_workitems/edit/300863#11941685
When the sharing option was set to _This establishment shares local governors with other academies in the trust_ then the duplicate would display.
This PR adds some code to catch the duplication and only display the local governor role.

**Major changes**
- `GovernorsGridViewModel` class updated with code to remove the equivalence checks.
- `GovernorGridViewModelTests` class updated with new unit tests.
- `GovernorControllerTests` class updated with a test that originally failed due to displaying 2 roles but after the code change now passes with an assertion on a single role being displayed.

A spike is going to be created for a developer to review and understand the why and what for the _RoleRequivalence_ function.

### DIFF
--- a/Web/Edubase.Services/Governors/Models/GovernorModel.cs
+++ b/Web/Edubase.Services/Governors/Models/GovernorModel.cs
@@ -2,9 +2,6 @@ using Edubase.Common;
 using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace Edubase.Services.Governors.Models
 {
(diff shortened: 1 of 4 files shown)


esde84 (approved):
Thank you for going through this, from a code point of view it seems to do what we need. I will let your testing be the proof of this (as best as you can test it). The unit test additions seem good too.

PR #864, opened 2026-04-17
Front end docs

martinwbailey:
Added C4 component diagram for the front end component. Added reference documentation for each sub component

Static web pages deployed here : https://dfe-digital.github.io/get-information-about-schools/


docs-site/layouts/partials/footer.html
```diff
@@ -0,0 +1,46 @@
...
+</footer>
+
+<% govuk_frontend_base = "https://cdn.jsdelivr.net/npm/govuk-frontend@6.0.0/dist/govuk" %>
+<script type="module" src="<%= govuk_frontend_base %>/govuk-frontend.min.js"></script>
+<script type="module">
+  import { initAll } from "<%= govuk_frontend_base %>/govuk-frontend.min.js";
+  initAll();
+</script>
+<script type="module">
+  import mermaid from "https://cdn.jsdelivr.net/npm/mermaid@11/dist/mermaid.esm.min.mjs";
```
rrees: Can we vendor this to avoid the external dependency initially? Also need to think about how we manage this for security updates.

docs-site/layouts/partials/key.html
```diff
@@ -0,0 +1,39 @@
+<h2 class="govuk-heading-m">Key</h2>
+<table class="govuk-table">
```
rrees: I realise this may have come from elsewhere but it feels like it should be a definition list

docs-site/layouts/ecf1-service-rules.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
```
rrees: Need to rename this layout (or maybe delete?)

docs-site/Gemfile.lock
```diff
@@ -0,0 +1,118 @@
+GEM
```
rrees, line 1: Let's enable checksums on the dependencies: `bundle lock --add-checksums`

documentation/service/front-end-component/reference/bulk-updates.md
```diff
@@ -0,0 +1,252 @@
...
+## How the web front end talks to the back end
+
+These flows all use typed API service clients rather than direct HTTP code in the controllers.
+
+The main services are:
+
+- `IEstablishmentWriteService`
+- `IGovernorsWriteService`
+
+Their Texuna-backed implementations send multipart uploads to endpoints such as:
```
rrees: Is the overview of Texuna usage explained elsewhere?

documentation/service/front-end-component/reference/change-history.md
```diff
@@ -0,0 +1,253 @@
...
+
+Important filter areas include:
+
+- Establishment fields
+- Establishment types
+- Group types
+- Suggester group
+- Approver group
+- Date range
+- date filter mode
```
rrees: Is there any significance to the casing differences here?

rrees (approved):

docs-site/layouts/partials/footer.html
martinwbailey: @rrees I tried vendoring Mermaid, but it tripped Gitleaks secret scanner, and the minified third-party bundle produced false positives. The only practical fix was to exclude that file from scanning, which felt like the wrong tradeoff because it creates a blind spot for a large vendored asset. I’ve therefore backed out the vendoring change and left the CDN-based import in place.

docs-site/layouts/partials/key.html
martinwbailey: @rrees Changed to use definition list

documentation/service/front-end-component/reference/change-history.md
martinwbailey: @rrees  No, This is consistent now

docs-site/layouts/ecf1-service-rules.html
martinwbailey: @rrees  This is redundant, and has been removed

documentation/service/front-end-component/reference/bulk-updates.md
martinwbailey: @rrees  Descriptive references to Texnua have been removed

docs-site/Gemfile.lock
martinwbailey: bundle lock --add-checksums enabled. Readme update to make note of this

docs-site/layouts/partials/footer.html
rrees: Okay, can we try and bring it in during the build phase so it is statically packaged on deployment?

martinwbailey: @rrees  Updated so that the Mermaid is pulled in during the Pages build and packaged into the site, instead of being loaded from the CDN in the browser

antk43 (approved):

PR #877, opened 2026-05-19
RK/Change-checking-est-type

skhurshidDFE:


Web/Edubase.Web.UIUnitTests/Controllers/ToolsControllerTest.cs
```diff
@@ -0,0 +1,143 @@
...
+            };
+
+            mockEstablishmentReadService.Setup(e => e.GetAsync(It.IsAny<int>(), It.IsAny<IPrincipal>()))
+               .ReturnsAsync(() => new ServiceResultDto<EstablishmentModel>(est));
+
+            var model = CreateBulkAcademiesViewModel();
+
+            var result = await controller.BulkAcademies(model, null, null, "search") as ViewResult;
+
+            Assert.NotNull(result);
```
antk43: We're asserting on a negative here and L61 but the test name includes the word `success` - is the test name wrong or could the test be split if the behaviour needs checking?

antk43 (commented):
Looks okay to me other than my comment. Let' get @anthony9853 to check and approve if happy.

Web/Edubase.Web.UIUnitTests/Controllers/ToolsControllerTest.cs
rajatkumar-dfe: Agree. that step was referenced from some other controller tests. I have now removed the step.

Web/Edubase.Web.UIUnitTests/Controllers/ToolsControllerTest.cs
```diff
@@ -0,0 +1,142 @@
...
+                mockEstablishmentReadService.Object,
+                mockEstablishmentWriteService.Object,
+                Mock.Of<ICachedLookupService>(),
+                Mock.Of<IClientStorage>(),
+                Mock.Of<ILocalAuthoritySetRepository>(),
+                mockEstablishmentDownloadService.Object);
+        }
+
+        [Fact]
+        public async Task BulkAcademies_Convert_AllowedType_Success()
```
antk43: Just looking into this in a bit more detail while we wait for @anthony9853.
The test name here suggests that an allowed type for the conversion results in a successful path. However, if we look at the assertions and debug through the controller we can see that the test is checking that with an allowed type the validation fails.
The test is therefore covering the search gate validation for an allowed type.

In the controller this test is only checking the gate of `AllowedEstablishmentTypeForBulkCreateAcademies`:
```c#
else if (!AllowedEstablishmentTypeForBulkCreateAcademies.Any(x => x == est.TypeId))
{
    ModelState.AddModelError(nameof(model.SearchUrn), "Please enter a valid URN");
}
```
The test is therefore not checking that conversion is possible nor that a success occurs; hence the ambiguity in the name.

An example test name that would be more suited for this test is something like 
`BulkAcademies_Search_AllowedType_ButFilteredTypesEmpty_AddsUrnError`

antk43 (approved):

PR #878, opened 2026-06-05
296879 Potential fix for code scanning alert no. 87: URL redirection from remote source

anthony9853:
Potential fix for [https://github.com/DFE-Digital/get-information-about-schools/security/code-scanning/87](https://github.com/DFE-Digital/get-information-about-schools/security/code-scanning/87)

In general, to fix untrusted URL redirection issues you must ensure that user input cannot arbitrarily control the redirect target. Here the base path is already fixed, but we still blindly forward the entire query string. The safest fix that preserves existing behavior as much as possible is to stop concatenating `Request.QueryString` directly, and instead rebuild a query string from specific, known-safe parameters on the current request and from the `viewModel`. This way, even if new parameters are added to requests elsewhere, they are not automatically forwarded through this redirect.

Concretely, we will change line 255–256 so that instead of:

```csharp
return Redirect(
    $"{Url.Action(ConstIndex, "GovernorSearch", new { area = "Governors" })}?{Request.QueryString}&{string.Join("&", viewModel.GovernorSearchModel.RoleId.Select(r => $"&{Areas.Governors.Models.GovernorSearchViewModel.BIND_ALIAS_ROLE_ID}={r}"))}");
```

we build a dictionary of allowed query parameters (for example: search terms, paging, filters) by pulling them from `Request.QueryString` by name, and then use an existing helper (if present) or `System.Web.HttpUtility.ParseQueryString` to construct a new query string. Since we are constrained to only modify the shown snippet and imports in this file, and we already see `QueryStringHelper` used elsewhere in the controller, the best fit is to use it to build the query string from selected keys rather than forwarding `Request.QueryString` wholesale. If other parts of the code rely on specific parameters in GovernorSearch, we should include those specific bindings (e.g., from `viewModel.GovernorSearchModel`) rather than an unbounded pass-through.

Because we have not been shown all the relevant parameter names, the minimal safe change that CodeQL will recognize is to remove `Request.QueryString` from the redirect and only include the governor role IDs (which are coming from the server-side model, not directly from the request). This slightly changes behavior by no longer forwarding arbitrary query parameters, but it preserves the primary functional intent of redirecting with the selected governor roles. If needed, additional explicit parameters can be added later using the same controlled pattern.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/Web/Edubase.Web.UI/Controllers/SearchController.cs
+++ b/Web/Edubase.Web.UI/Controllers/SearchController.cs
@@ -252,8 +252,10 @@ namespace Edubase.Web.UI.Controllers
 
                         if (viewModel.SearchType.OneOfThese(eSearchType.Governor, eSearchType.GovernorReference, eSearchType.GovernorAll))
                         {
-                            return Redirect(
-                                $"{Url.Action(ConstIndex, "GovernorSearch", new { area = "Governors" })}?{Request.QueryString}&{string.Join("&", viewModel.GovernorSearchModel.RoleId.Select(r => $"&{Areas.Governors.Models.GovernorSearchViewModel.BIND_ALIAS_ROLE_ID}={r}"))}");
+                            var baseUrl = Url.Action(ConstIndex, "GovernorSearch", new { area = "Governors" });
+                            var roleIdParams = string.Join("&", viewModel.GovernorSearchModel.RoleId.Select(r => $"{Areas.Governors.Models.GovernorSearchViewModel.BIND_ALIAS_ROLE_ID}={r}"));
+                            var redirectUrl = string.IsNullOrEmpty(roleIdParams) ? baseUrl : $"{baseUrl}?{roleIdParams}";
+                            return Redirect(redirectUrl);
                         }
 
                         throw new NotSupportedException($"The search type '{viewModel.SearchType}' is not recognised.");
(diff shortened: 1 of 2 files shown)


Web/Edubase.Web.UI/Controllers/SearchController.cs
```diff
@@ -252,8 +252,10 @@ public async Task<ActionResult> IndexResults(SearchViewModel viewModel)
 
                         if (viewModel.SearchType.OneOfThese(eSearchType.Governor, eSearchType.GovernorReference, eSearchType.GovernorAll))
                         {
-                            return Redirect(
-                                $"{Url.Action(ConstIndex, "GovernorSearch", new { area = "Governors" })}?{Request.QueryString}&{string.Join("&", viewModel.GovernorSearchModel.RoleId.Select(r => $"&{Areas.Governors.Models.GovernorSearchViewModel.BIND_ALIAS_ROLE_ID}={r}"))}");
+                            var baseUrl = Url.Action(ConstIndex, "GovernorSearch", new { area = "Governors" });
+                            var roleIdParams = string.Join("&", viewModel.GovernorSearchModel.RoleId.Select(r => $"{Areas.Governors.Models.GovernorSearchViewModel.BIND_ALIAS_ROLE_ID}={r}"));
+                            var redirectUrl = string.IsNullOrEmpty(roleIdParams) ? baseUrl : $"{baseUrl}?{roleIdParams}";
+                            return Redirect(redirectUrl);
```
adamp10, line 262: Just checking that RoleIds are the only parameters needing to be carried across to GovernorSearch, ie no other search/filter criteria need to be preserved?

adamp10 (approved):

PR #879, opened 2026-06-05
Added PRU in allowed list for establishment conversion

rajatkumar-dfe:
Identified PRU being a valid candidate but still not present in list of allowed establishments. Made this change on c# side to bring that.


adamp10 (commented):
This is probably okay to approve for now but some thought needs to go into the design since the validation/filtering now appears to be a mix between front-end and backend and it's a bit confusing. Also, the add/edit paths still use the full list when calling the backend filtering/validation so that's different from the search path now.

Unit tests might also need to be added for coverage.

adamp10 (approved):

